PR #4, opened 2018-04-21
Fix NULL pointer dereferences and improve compatibility with COM

JohanSmet:
COM AddRef/Release return the new reference count. Changing FAudio_AddRef/Release to do this as well prevents a duplicated reference counter in the C++ wrapper.

XAudio2Create allocates the XAudio implementation and initializes the engine. The COM-wrapper needs to be able to allocate without initializing in CreateInstance. I saw no other solution than to split off this functionality from FAudioCreate to another public function (FAudio_Construct).

### DIFF
--- a/src/FAudio.c
+++ b/src/FAudio.c
@@ -1206,7 +1206,7 @@ uint32_t FAudioSourceVoice_FlushSourceBuffers(
 	/* Go through each buffer, send an event for each one before deleting */
 	while (entry != NULL)
 	{
-		if (voice->src.callback->OnBufferEnd != NULL)
+		if (voice->src.callback != NULL && voice->src.callback->OnBufferEnd != NULL)
 		{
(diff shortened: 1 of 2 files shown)


src/FAudio_internal.c
```diff
@@ -525,7 +525,7 @@ static void FAudio_INTERNAL_MixSource(FAudioSourceVoice *voice)
 	}
 
 	/* Filters */
-	if (voice->flags & FAUDIO_VOICE_USEFILTER)
+	if (voice->flags & FAUDIO_VOICE_USEFILTER && voice->filterState != NULL)
```
flibitijibibo: This can be removed, seems smarter to just allocate on `USEFILTER` right away:

https://github.com/flibitijibibo/FACT/commit/f6ef20b5294f9451695ec8090b8aaf3000104f8b

PR #7, opened 2018-05-01
Fix compilation on VS2010

JohanSmet:
I noticed that the latest FAudio doesn't compile with VS 2010 anymore while updating the COM-wrapper to work with VS 2010.


flibitijibibo:
Why does GitHub not default to watching your own repositories...?

Anyway, merged both patches, sorry this didn’t get done sooner!

PR #8, opened 2018-05-06
F3DAudio "beta"

lamarqua:
All the matrix computation cases are handled, including multi-channel emitters, REDIRECT_TO_LFE / ZERO_CENTER, cones and custom curves.
Done some manual testing and everything was giving good values (with differences due to the fact that I lerp everywhere) and out of the right speakers, so it's a good candidate for a field test.
Non-implemented features (LPF / Reverb) have a minimal implementation that sets values, so an uninitialized input structure wouldn't come out with garbage. As such it can be tested with the COM wrapper as well and nothing should blow up.


lamarqua:
ACK. Indeed we're probably the first. Judging by my google searches, I don't get the impression that many people used X3DAudio directly, either. (Seems like most used it through XACT?)

What do you mean by re-sort exactly? I tried to keep stuff grouped by locality of usage, although it might not be clear at first glance.

> Someone needs to be able to read this top-to-bottom and understand it the very first time; we may very well be the first to publish a public implementation of this API so it's important we get it right, especially the stuff we know we got wrong.

I'll try to write more detailed comments, but I'd be open to suggestions as to what could be made clearer. I don't think it'd be possible for someone unfamiliar with the API to understand the code, no matter how detailed the comments would be. (And in particular, someone who hasn't read the MSDN docs _thoroughly_ and played with the API, which is what I did.) The 3D maths isn't complicated but conveying information about the calculations is tough when all you have is ASCII. (Diagrams only go so far.) The docs also are often very vague about the exact behaviour, which is why I had to use so many curve plots to figure out exactly what happened.

I agree though that I'll need to comment more what is accurate and where the behaviour deviates from X3DAudio. 

I was also planning on doing some kind of writeup of the whole effort for my blog (by request of friends of mine), but that is at the very least a month or two away from now. 

I'm flying tomorrow and I'll be really really busy until the 20th, but I'll do my best to do at least the easy stuff asap. (Lines, asserts, etc.)

I hope we can at least move in parallel on that -- I fix the stylistic and "real users" test ;-) I'm eager to see how my code holds up in the real world. (I'll keep my development in the same branch in any case.)

flibitijibibo:
No worries on the timeline, for now what matters is that it works, and once we know it works I can help organize and point out things that need to be documented. I don't think we'll have to visualize much but naming any sort of formulas used, no matter how simple, will be helpful to those who don't have strong math skills (read: people like me who normally just make programs fast without understanding the science behind them).

I've let @bartwe know of this patchset, so he'll be in control for the most part until I can get out of the Switch Hole(TM). Johan may have some feedback too depending on what X3DAudio data is available to us.

lamarqua:
Sounds good to me. 👍

bartwe:
K did some testing, some of the following may be FNA, not FACT, issues.

- FAudio.cs F3DAudioCalculate has pDSPSettings as out param, should be ref
- F3DAUDIO_CALCULATE_MATRIX is commented out ;)
- Early test of F3DAUDIO_EMITTER.pVolumeCurve seems to work, will do some more testing of it.

On the FNA side

- dev.DSPSettings.SrcChannelCount and dev.DSPSettings.DstChannelCount need to be initialized before call to F3DAudioCalculate
- Call to F3DAudioCalculate doesn't pass F3DAUDIO_CALCULATE_MATRIX

Unsure which side

- Left/Right are flipped

Otherwise seems to be functional, did not test for how close it is to xna (i don't have an xna baseline of staxel anymore)

flibitijibibo:
`out/ref` param keywords have been fixed, as has Src/DstChannelCount assignments. This pull request will need to uncomment the following pieces:

https://github.com/flibitijibibo/FACT/blob/master/src/FACT3D.c#L130
https://github.com/flibitijibibo/FACT/blob/master/src/FACT3D.c#L150

On FNA's end, the following parts need to be uncommented (this needs to be done manually until this PR is merged):

https://github.com/FNA-XNA/FNA/blob/fact/src/Audio/SoundEffectInstance.cs#L230
https://github.com/FNA-XNA/FNA/blob/fact/src/Audio/SoundEffectInstance.cs#L294

Left/Right flipping I may need an example, it's possible that the matrix coefficients in Calculate are flipped by accident...?

lamarqua:
If you provide an example of a call to F3DAudioCalculate with the audio
flipped I can check it against the reference implementation :)

JohanSmet:
I did a few tests with the COM-wrapper and had no big issues. I only have a simple 2-speaker setup ATM so I can only test simple spatialization.  I tried two applications:
- in Capsized the sounds come from the correct speaker. I can't tell the difference between your implementation or Microsoft's.
- the june 2010 dxsdk example XAudio2Sound3D works great when the Inner Radius is disabled. With the inner radius enabled sounds seems to come from the wrong speaker when the audio source is inside the inner radius of the listener. Here's a [video](https://youtu.be/QhxmMhmQ3Pc) demonstrating this and this [video](https://youtu.be/x4h8vzBAuBQ) shows how the MS implementation behaves in the same situation.

lamarqua:
I will have a look at it!

src/F3DAudio.c
```diff
@@ -28,174 +28,1338 @@
...
+enum {
+    DIFFUSION_SPEAKERS_ALL = 0,
+    DIFFUSION_SPEAKERS_MATCHING = 1,
+    DIFFUSION_SPEAKERS_OPPOSITE = 2
+};
+
+/* Adrien: determined experimentally; this is the midpoint value, i.e. the value at 0.5 for the matching speakers, used for the standard diffusion curve. Note: it is SUSPICIOUSLY close to 1/sqrt(2), but I haven't figured out why. */
+#define DIFFUSION_LERP_MIDPOINT_VALUE 0.707107f
+
+static void ComputeInnerRadiusDiffusionFactors(float radialDistance, float InnerRadius, float *diffusionFactors)
```
flibitijibibo: Would this function be the one where InnerRadius affects the speaker output? Just trying to narrow this down in case someone wants to step through this.

lamarqua: Yes, that is the most probable one. I am still super busy until Sunday but I promise to look at it next week.

flibitijibibo: No worries, as long as we can narrow it down - Bart and Johan may be able to prod at this if they have time first.

flibitijibibo:
Totally blanked on this, anything new on this end? I've been submerged in FACT accuracy/memory fixes so I haven't been able to play around with the InnerRadius math.

lamarqua:
I've been super busy this week so I haven't had a chance to take a look at
it yet.

flibitijibibo:
Going to do a mass cleanup of this PR today and send it back to this Git branch, so that it's easier for everyone to read and debug in case someone else needs to. Most of it will be style but I'm going to strip out a lot of the debug stuff too. Will separate into two commits, one for non-tabbing cleanup and another for tabbing cleanup.

flibitijibibo:
https://github.com/lamarqua/FACT/pull/1

flibitijibibo:
One more pull request to uncomment the FACT3D stuff. The FNA SoundEffectInstance 3D calls still need to be uncommented but for XACT data this should work out of the box now:

https://github.com/lamarqua/FACT/pull/2

flibitijibibo:
Made a quick update to FNA for those testing with that, here are the new spots to uncomment if you're using SoundEffectInstance:

https://github.com/FNA-XNA/FNA/blob/fact/src/Audio/SoundEffectInstance.cs#L211
https://github.com/FNA-XNA/FNA/blob/fact/src/Audio/SoundEffectInstance.cs#L282

flibitijibibo:
As soon as I posted that I had the shocking revelation that maybe if I just set the matrix coefficients to something basic we don't have to have comment/uncomment magic going on:

https://github.com/flibitijibibo/FACT/commit/99a5ea5c749582c3b87b3f03ea4d6e9f4e41c420
https://github.com/FNA-XNA/FNA/commit/ab0661339c3586c00093d49d45a64e173b20702b

One more pull request as a result, to deal with the conflict...

https://github.com/lamarqua/FACT/pull/3

flibitijibibo:
After today's post spamming and a big pile of commits I can now happily say that this branch works with FNA perfectly! 3D seems to work great for FNA's purposes and surround sound works fine as well. Right now you have to hardcode a 5.1 config because SDL doesn't give this to us...

https://github.com/flibitijibibo/FACT/blob/master/src/FAudio_platform_sdl2.c#L417

... but that's an SDL problem, not an FAudio or F3DAudio problem.

As soon as this InnerRadius thing is fixed and @bartwe's verified the extra stuff on their end, all that will be left is documentation, then we can FINALLY call this done.

lamarqua:
Happy to hear that it seems to work fine.
I'm investigating the audio issue on my end, but I am now currently stuck in a little bit of configuration and setup problem as I try to get a virtual machine to behave somewhat like the windows machine I left behind when moving here. (I only have a Mac laptop right now.) It's a little bit hellish, to be honest, but I should be able to run and debug code before tomorrow.

lamarqua:
I have a barely usable setup to test and I can confirm the speaker reversal exists. (And there is some other funky stuff happening as well.)
I'm investigating it right now.

FYI, your style cleanup broke the distance curve code: https://github.com/lamarqua/FACT/commit/fa5e7515ab6611f907f9ee6f763c95ec0c5306c9

I don't have the courage to triple check the massive changes in detail right now, but maybe you remember maybe some of the places where you did changes like that (moving things in and out of conditions)?

lamarqua:
(The commit I linked to reversed the code to what was previously done.)

flibitijibibo:
I believe that's the only one I did, my mistake there... usually when I see `++` instead of `+= 1` I assume there's a place I can trim out a line. If there are any more I'm forgetting, it'll be where `++`/`--` are used.

lamarqua:
🎉 Found the bug! 🎉 

It's a dumb mistake that I didn't catch because it only happens in the stereo case, not in the more complex configurations.
When `InnerRadius` is set, the behaviour is defined as such: the sound is "diffused" according to some distribution, depending on the distance. 
That's what `ComputeInnerRadiusDiffusionFactors` computes: it gives you 3 floats, which I call the diffusion factors, each of which represent how much you should apply a certain kind of diffusion: EQUAL_ENERGY is dividing the total energy (the one computed after distance attenuation and stuff) by the number of speakers, and giving each of them that amount, MATCHING_SPEAKERS is giving some energy to the speakers that are closest in azimuth to the ones in the current speaker configuration, and OPPOSITE_SPEAKERS bleeds some of the signal to the opposite speakers of the matching ones. (This behaviour is somewhat explained [here](https://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.x3daudio.x3daudio_emitter(v=vs.85).aspx))
This is a "triple law" (which I plot-reversed and approximated; I have the curves somewhere if you're interested), that is for some values, all of these factors may be non-zero, with the added fact that they always sum to 1.

In the stereo case, matching speakers and opposite speakers are the same pair! (It does not happen in any of the other configurations except MONO and STEREO 2.1, because they all have more than 4 speakers. I had only tested on the more complex configurations... my mistake.)

What happens is given the diffusion factors, `ComputeEmitterChannelCoefficients` would set the right matrix coefficients, in order: first EQUAL_ENERGY, then MATCHING_SPEAKERS, then OPPOSITE_SPEAKERS. But in the stereo case, OPPOSITE_SPEAKERS would overwrite the values first set by the code which previously set MATCHING_SPEAKERS.

The fix is simple: since I zero the coefficients memory earlier in the code, instead of using `=` when I set the matrix coefficients, I can just use `+=` and not have to worry about matching and opposite speakers being the same in the STEREO case.

I am reasonably confident (95%) that this should fix the problem and not introduce any more bugs, but for sure there might still be other corner cases like that in the code base that I didn't catch.

I'm going to be fairly busy in the next 4 to 8 weeks because I'm on a job search right now; but I'd like to do whatever you feel needs to be done in whatever spare time I have so that we can close the bounty ASAP.

flibitijibibo:
Beautiful! That _should_ be the last of the code changes, but I need @bartwe to verify this one more time.

The very last thing I'll do is do a final style pass for all the minute details, as well as a documentation pass where I'll essentially be dropping in `/* FIXME: Explain this! */` blocks wherever I get confused. Once all the doc FIXMEs are done and Bart verifies this on his end, this can be merged and we can finally pay this thing out!

flibitijibibo:
Just got word from Bart that this seems to work for Staxel, so let's get this documented and we'll call this job done.

(Quick reminder for the FIXME DOCS PR: https://github.com/lamarqua/FACT/pull/4 )

lamarqua:
Superb. I'll work on it this week.

lamarqua:
I've added comments to all the places you mentioned as FIXME. I'm not sure I can do much better than that without going into full blown math explanations (which is kinda awkward in source code comment...) or repeating what is explained in more detailed and more clearly in the MSDN. Let me know what you think.

flibitijibibo:
Just took a look, everything looks great now! @bartwe, we need one final test, and this patch in its current form works I'll merge this in.

@lamarqua, is the e-mail address used in the Git log good for contacting re: bounty payouts and such?

lamarqua:
Wow that was quick! Yes my email is up to date. 
Note that I still plan to do a write up for my blog of the whole REing process, but that won't happen until a couple months.

flibitijibibo:
Verified this one more time on my end, including a VS2010 fix:

https://github.com/lamarqua/FACT/pull/5

@bartwe, have a couple minutes to check this out one more time?

flibitijibibo:
Looks like Bart's out this weekend, so I'll merge this based on it working for FNA/FACT's standard behavior and it working with Adrien's test data. If the cone stuff doesn't work for whatever reason, we can revisit this.

PR #9, opened 2018-05-08
Clear voice->src.bufferList when all buffers are removed

JohanSmet:
This fixes a potential infinite loop in voice->src.bufferList after a FlushSourceBuffers and SubmitSourceBuffer combo.


flibitijibibo:
Oh if only Valgrind could have found this...

Thanks much!

PR #11, opened 2018-05-16
XAudio2 Wine DLLs

JohanSmet:
Implements https://github.com/flibitijibibo/flibitBounties/issues/4


flibitijibibo:
Finally got to try this out in depth and everything works incredibly well! This is really really good work, thanks for taking this on.

The only two things I noticed are based on one common thing: SDL2.dll doesn't get copied into the output folder, so `regsvr32` fails silently because of the classic "module could not be found" message, which is Windows trying to tell us SDL2 is what's actually missing...

It may be worth documenting that in the README for those who intend to take the DLLs out of the VC output folder and ship them somehow, and for RegisterDLLs maybe we can add a line to copy from the SDL2-devel folder to do that part automatically for Windows. Not sure how much it matters for the average Wine hacker but if we can do that there too, could be helpful!

Once that's fixed that'll be it, I think. I could probably dig super deep into style things but you've spent enough time on this and I can always take a personal day to do silly things like that.

JohanSmet:
Ah, that's a good remark. I had SDL2.dll in the %PATH% somewhere and didn't think about it anymore.

The VC RegisterDLLs project and the mingw Makefile both copy SDL2 to the output directory now. And the wine setup script also links it into the wine-prefix.

The README has been extended with a small note about redistributing the DLLs.

flibitijibibo:
Tried it out, works great, another job well done!

PR #12, opened 2018-06-04
Reverb

flibitijibibo:
This is just a GitHub tracker for the reverb work - easier to have here than to spam Johan with a bunch of barely-threaded e-mails in a row!


src/FAudioFX.c
```diff
@@ -137,30 +138,267 @@ static FAPORegistrationProperties ReverbProperties =
...
+		/* early reflections */
+		float early = FAudioFXFilterAllPass_Process(reverb->apf_in[0], delay_in);
+		early = FAudioFXFilterAllPass_Process(reverb->apf_in[0], early);
+
+		early = early * reverb->early_gain;
+
+		/* reverberation */
+		float revdelay = FAudioFXFilterDelay_Process(reverb->reverb_delay, early);
+		float comb_out = 0.0f;
+		float gain = 0.125f;
```
flibitijibibo: I notice that if I set this to 1.0f it sounds pretty good with our test case, but I don't know if there's something important to this value...

flibitijibibo:
@JohanSmet Do we have a checklist for what's left before this is ready? For Dust this still sounds great but I dunno if there's anything we need to cover before merging this.

JohanSmet:
Hmm, it seems GitHub does not send a notification when somebody else creates a PR for a branch you're working on (or I missed it somehow?). Didn't notice this existed before your last comment...

Most of the items on my todo list have been completed. Three are still open:
1. Surround support: the reverb effect should be able to 'upgrade' from 1/2 channel input to 5.1 output.
   - should be straight forward to extend the current stereo path to more channels and add an extra delay (for parameter 'RearDelay') and potentially some extra attenuation.
   - requires some modifications to FAudio before this can be implemented/tested:
      - check and use the 'OutputChannels' parameter from the EffectDescriptor to potentially override the channel count of the voice (and fix any side effects)
      - validate / reject modifications of the effect chain
2. Tweak the reverb effect to sound and behave more like the original XAudio2 reverb
    - this is a somewhat open ended todo item ... we can change the filter topology and tweaks delays but without knowing the layout of the Princeton Reverb it's all guesswork and trial and error
    - I think the best thing to do is to test as much real work use cases as possible and only tweak the algorithm when there's a combination of parameters that sound really bad.
3. Go through the code one final time and add missing asserts / comments

I can do item 1 (the reverb part) and item 3 tomorrow (maybe overflowing to Thursday).

flibitijibibo:
That works for me! Definitely don't worry about 2 unless you have specific data that doesn't match; because we have no reference point for both the standard and proprietary reverb formats this can remain totally open-ended until we come across data that gives us new parameters to work with. That'll ultimately be the responsibility of the FNA community to test when the FAudio branch goes into beta.

JohanSmet:
I've added 5.1 support to the reverb effect and fixed up some things in a last pass through the code. The surround output has not been tested with FAudio but I did hook up FAudioFXReverb with the native XAudio2 on windows and it seems to work fine there.

There were a few changes required to make it work with XAudio2 (e.g. XAudio2 keeps calling the effect with buffers marked as XAPO_BUFFER_SILENT when there are no source buffers left and the silent buffers are filled with garbage).  

There's one issue that I don't know how to fix cleanly without breaking the FAudio path. XAudio2 expects IXAPO::GetRegistrationProperties to allocate the required memory with XAPOAlloc (a macro that resolves to CoTaskMemAlloc). Otherwise the Windows RTL freaks out when XAudio releases the memory with CoTaskMemFree. Because this is such an unlikely use scenario, I just added a note to the code (https://github.com/flibitijibibo/FACT/pull/12/commits/cc1fabbeec4e81103ad1d09f63f279d9d045646d) and moved on.

The branch should be in good shape to merge now.

flibitijibibo:
That XAPOAlloc thing is... something special. I think the note is good enough, the odds of the two overlapping in any serious way is, as you say, basically zero.

Will give this a go on the 5.1 setup and test a couple more scenarios over the weekend!

flibitijibibo:
Gave this a go on the 5.1 setup and there seems to be only one real accuracy issue, which is that rear reflections don't seem to happen. Sadly I don't have a way to record XAudio2's version, but it basically sounds the same as FAudio except the rear speakers get an extra reflection depending on the rear parameters.

Luckily this can be tested even without a real 5.1 setup; run `SDL_AUDIODRIVER=disk ./FAudioReverbDemo` and when 5.1 is enabled it'll write float32 6-channel raw PCM that can be imported by Audacity (File > Import > Raw Data). From there you should be able to see the rear channels.

There also seem to be some other issues on Windows but I think that's just our Windows audio support being a bit crap... Linux channel mapping is totally fine.

JohanSmet:
Ah, right. I think I know what's happening. I only added 5.1 support to the reverb effect itself. I didn't change FAudio to handle a FAPO that changes the number of channels. Right now FAudio ignores the OutputChannels fields of the EffectDescriptor and you're only a getting mono/stereo reverb even when FAudioReverbDemo is running in 5.1 mode.

Properly handling a effect that changes the channel count will require some changes to FAudio. From what I can tell now (I am not that familiar with this part of FAudio):
- changing the resampleCache size after the effect runs
- this also means no in-place processing of the effect
- check out the effect on SetOutputMatrix
- valid channel count checking has to change depending on if it's before or after the effect.
- ... ?

I have some time this weekend to look into this, if you want. But maybe this should be a separate branch/PR because it changes core FAudio and not FAudioFX?

flibitijibibo:
Ah, I see what's up - so here's what we have to do:

- Validate [SetEffectChain](https://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.ixaudio2voice.ixaudio2voice.seteffectchain(v=vs.85).aspx) input
    - Add a `float *effectChainCache`, alloc if in-place processing is unsupported
- Change [SetOutputMatrix](https://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.ixaudio2voice.ixaudio2voice.setoutputmatrix(v=vs.85).aspx) to validate `SourceChannels` based on the final channel count
- Add a return value to `INTERNAL_ProcessEffectChain` which returns `resampleCache` _unless_ `effectChainCache` is used (i.e. an effect could not do in-place processing), and whatever the return value is gets used for the final mix

The good news is that all of the work is almost entirely contained within SetEffectChain, SetOutputMatrix, and ProcessEffectChain, with maybe a couple tweaks to the final mix loop.

You're right though, that doesn't even touch the FAudioFX stuff... I'll go ahead and get this merged today since FACT's case is covered anyhow.

JohanSmet:
Ah, that doesn't look as complicated as I initially feared :-)

Thanks for the merge!

PR #18, opened 2018-07-26
Wine fixes

aeikum:
These changes are motivated by getting FAudio to build in Wine's build system, and fixing a few of the problems that Wine's tests turned up.

In particular, I think "Update FAudioVoiceDetails" and "Advance buffer offset" need careful review, as they're significant changes that could impact existing clients.

More to come ;)


flibitijibibo:
Testing locally everything seemed great! The buffer offset stuff was always really scary so having a second pair of eyes on that makes me feel a whole lot better about it. I tried with a bunch of streaming XACT MSADPCM data and it worked perfectly fine, and that's probably the nastiest kind of test data you can find :P

PR #20, opened 2018-07-27
Fapo

aeikum:
These changes allow us to implement FAPO objects external to FAudio.


flibitijibibo:
Always figured FAPO's layout would cause a fuss some day...

FNA tests cases are running fine, so I'm good with this!

PR #28, opened 2018-08-24
Fix COM XACT 3.4 (and earlier)

JohanSmet:
I was too slow to help with the visual studio projects for the xact com wrapper :-)

Tested it with Capsized on windows and it crashed in the COM initialization. Turns out the interface changed slightly between version 3.4 and 3.5 of XACT. With this change Capsized gets through the COM initialization fine. But unfortunately it fails in FACT_INTERNAL_ParseAudioEngine() because it uses an older content version (45).

Verified that PB Winterbottom still works, didn't run any other tests.


flibitijibibo:
Yeah, Capsized Windows is still on XNA 3.1... I wonder what changed between 45 and 46? Anyway, thanks for checking that out!

PR #35, opened 2018-10-03
Minor tweaks

benoit-pierre:
* fix typo in the documentation 
* tweak wrapper cross-compilation scripts: use `return` instead of `exit` to avoid killing the shell on error handling when sourced.


flibitijibibo:
Not much to see here, thanks for the quick patch!

PR #41, opened 2018-10-07
Bare-bones CMakeLists

cybik:
This is *extremely* bare-bones and covers only the "bring library into codebase that has an XAudio2-based implementation already". But it actually works!


flibitijibibo:
What is this meant to do that the Makefile/visualc folders don’t? Is this being included in another specific project? In that case I would add the CMake stuff to that project and add this as a submodule untouched.

cybik:
This would be useful for CMake, SDL2-based projects that would want to use the whole FAudio component if the engine already has an XAudio2 implementation, for example in the case of an engine that already has a full XAudio2 implementation and would like to reuse that code on Linux and OSX.

I'm using this to compile FAudio into a shared library for a game engine I'm contributing to, and I'm using CLion and a CMake-based structure on Linux and OSX. If this is completely outside of the scope of this library, feel free to nope this PR out.

flibitijibibo:
Ah, gotcha. Yeah in this case I would add it to the engine CMake project and add this as a submodule. I would even recommend static linking so it only links to SDL2 externally.

One thing that’d definitely be nice to have in our repo is fixes to the cpp work to make it a drop-in wrapper for XAudio2; the COM wrapper source is actually super close to being a portable C++ wrapper but there’s some gunk like the com_utils stuff that gets in the way. If you can make our headers/source work with that engine without modifying the engine, that’d be a huge feature for FAudio and I bet it would make the other engine maintainers really happy to have that much less code on their end.

cybik:
I basically took their XAudio2 code, removed the COM gunk and adapted the mutex stuff to use SDL2 code. The rest was REALLY just renaming the calls; almost feels like a drop-in.

flibitijibibo:
It definitely cones close! The next stuff for us is making the renaming unnecessary. It’s usually not much more than busy work in practice, but no work is always better!

flibitijibibo:
As a heads up, `master` now uses CMake instead of raw Makefiles!

PR #47, opened 2018-10-18
GCC utils fix

JohanSmet:
ecadcb7c652b3a90bd78cb1e3091b33a83a198ff broke compilation of utils on Linux. At least it did for me with GCC 7.3.


flibitijibibo:
Oh dang, I knew that existed for more than VS... sorry about that!

PR #52, opened 2018-11-05
Added Xcode-iOS Project

TheSpydog:
Builds FAudio static libraries for iOS and tvOS.


flibitijibibo:
Ran xcodebuild on this and it works! I only have 2 notes and they're both easy, I hope:

1. XNA_Song.c is missing
2. Let's aim for the lowest possible version that Xcode will allow - the output from xcodebuild suggests we can do 8.0?

Other than that I have no idea what I'm doing, so I'll take the rest at face value.

flibitijibibo:
Tried out the 8.0 update, it seems we depend on an ARMv8 NEON extension, so the lowest we can go is iOS 11. This removes 32-bit ARM from our supported list, but since Apple cut them off already I'm not too heartbroken about that...

The tvOS build was okay, so if targeting iOS 11 still works for you then that's what we'll go with.

PR #55, opened 2018-11-07
Import wine tests

aeikum:
This adds a couple of behavior changes to pass the tests, and then imports the Wine test suite in a new subdir, `tests/`. Its makefile supports building a Windows executable, to verify the behavior of Microsoft's implementation, and includes a compat header that mostly #defines XAudio2 symbols into FAudio symbols, to verify FAudio's behavior against the tests.

Note that the CLSIDs commit will change the default CLSID value in the FAPO's properties struct.

### DIFF
--- a/src/FAPOFX_echo.c
+++ b/src/FAPOFX_echo.c
@@ -31,6 +31,50 @@
 
 const FAudioGUID FAPOFX_CLSID_FXEcho =
 {
+	0x5039D740,
+	0xF736,
+	0x449A,
+	{
+		0x84,
+		0xD3,
+		0xA5,
+		0x62,
+		0x02,
+		0x55,
+		0x7B,
+		0x87
+	}
+};
+
+static FAPORegistrationProperties FXEchoProperties =
+{
+	/* .clsid = */ {0},
+	/* .FriendlyName = */
+	{
+		'F', 'X', 'E', 'c', 'h', 'o', '\0'
+	},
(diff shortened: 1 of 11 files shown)


aeikum:
I should add that Masanori Kakura and I are the only major contributors to these tests. I asked Masanori via email if he is OK with re-licensing his tests for this project and he replied that he was.

src/FAPOFX_echo.c
```diff
@@ -46,9 +90,9 @@ const FAudioGUID FAPOFX_CLSID_FXEcho =
 	}
 };
 
-static FAPORegistrationProperties FXEchoProperties =
+static FAPORegistrationProperties FXEchoProperties_LEGACY =
 {
-	/* .clsid = */ {0},
+	/* .clsid = */ FAPOFX_CLSID_FXEcho_LEGACY,
```
flibitijibibo: I’d have to double check tonight (currently away from a PC), but I believe MinGW complains about this clsid initialization. I totally forget why! For reference I’m building with mingw-w64 from Fedora 28.

Everything else lgtm!

aeikum: Ah I was wondering why that memcpy was there, it seemed obviously wacky. I can change it back.

aeikum:
OK, I re-wrote the branch with that fixed up.

flibitijibibo:
Great, thanks to both of you for the tests!

PR #57, opened 2018-11-09
Debug logging

aeikum:
This adds some debug logging functions and a couple useful debug lines around creating source voices and processing their buffers. I don't know if it will work well on your non-Linux platforms, maybe they should be made part of the platform API?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ endif
 ifeq ($(FAUDIO_RELEASE),1)
 	CFLAGS += -O3
 else
-	CFLAGS += -g -Wall -pedantic
+	CFLAGS += -g -Wall
 endif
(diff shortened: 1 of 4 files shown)


flibitijibibo:
We might be able to map this to SDL_Log, but as far as the internal workings go, this may be a good time to look at DEBUG_CONFIGURATION:

https://docs.microsoft.com/en-us/windows/desktop/api/xaudio2/ns-xaudio2-xaudio2_debug_configuration

In addition to programs asking for it via SetDebugConfiguration (I've hit this a couple times with some Steam games), we could also check environment variables to forcibly enable certain flags/features.

flibitijibibo:
Closing in favor of #59, though what's been written may still be useful for that issue.

PR #58, opened 2018-11-11
fix 64 bit mingw build on linux

GloriousEggroll:
Still something up with 32 bit and XAPObase, but this at least fixes the 64 bit build so it's usable


cpp/Makefile
```diff
@@ -189,10 +225,10 @@ $(XAUDIO27_TARGET) : directories $(XAUDIO27_OBJ) $(XAPO_TARGET)
 	$(CXX) $(CXXFLAGS) -shared -o $@ $(XAUDIO27_OBJ) $(XAPO_TARGET) $(LDFLAGS)
 
 $(XAUDIO28_TARGET) : directories $(XAUDIO28_OBJ) $(XAPO_TARGET)
-	$(CXX) $(CXXFLAGS) -shared -o $@ $(XAUDIO28_OBJ) $(XAPO_TARGET) $(LDFLAGS) -lsetupapi
+	$(CXX) $(CXXFLAGS) -shared -o $@ $(XAUDIO28_OBJ) $(XAPO_TARGET) $(XAPOFX15_TARGET) $(LDFLAGS) -lsetupapi
```
flibitijibibo: For these two targets, add XAPOFX.cpp to XAUDIO2_SRC_x28. CreateFX is actually inside the XAudio2_8 dll itself, so this isn't quite the same.

cpp/Makefile
```diff
@@ -240,15 +291,19 @@ directories:
...
 		cp $(FAUDIO_FFMPEG_PREFIX)/bin/avutil*.dll $(BUILD_DIR);	\
+		cp $(FAUDIO_FFMPEG_PREFIX)/bin/swresample*.dll $(BUILD_DIR);	\
 	fi
 
 clean:
 	rm -f $(XAUDIO20_TARGET) $(XAUDIO21_TARGET) $(XAUDIO22_TARGET) $(XAUDIO23_TARGET) \
 	      $(XAUDIO24_TARGET) $(XAUDIO25_TARGET) $(XAUDIO26_TARGET) $(XAUDIO27_TARGET) \
-	      $(XAUDIO28_TARGET) $(XAUDIO29_TARGET) $(X3DAUDIO_TARGET) $(XAPO_TARGET) \ \
+	      $(XAUDIO28_TARGET) $(XAUDIO29_TARGET) $(X3DAUDIO_TARGET) $(XAPO_TARGET) \
+              $(XAPOFX11_TARGET) $(XAPOFX12_TARGET) $(XAPOFX13_TARGET) \
```
flibitijibibo: Extra couple spaces?

cpp/scripts/wine_setup_native
```diff
@@ -78,12 +78,16 @@ install_dll xactengine3_5 1
...
-install_dll xapofx1_5 0
+install_dll xapofx1_1 1
+install_dll xapofx1_2 1
+install_dll xapofx1_3 1
+install_dll xapofx1_4 1
+install_dll xapofx1_5 1
 
 ln -sf "$dll_path/FAudio.dll" "$wine_path/FAudio.dll"
 ln -sf "$dll_path/SDL2.dll" "$wine_path/SDL2.dll"
+ln -sf "$dll_path/avcodec-58.dll" "$wine_path/avcodec-58.dll"
```
flibitijibibo: For the FFmpeg files, check to see if they exist first before linking them.

cpp/scripts/wine_setup_native
```diff
@@ -63,11 +63,11 @@ install_dll xaudio2_7 1
 install_dll xaudio2_8 1
 install_dll xaudio2_9 1
 
-install_dll x3daudio1_3 0
-install_dll x3daudio1_4 0
-install_dll x3daudio1_5 0
-install_dll x3daudio1_6 0
-install_dll x3daudio1_7 0
+install_dll x3daudio1_3 1
```
flibitijibibo: Now that we're installing every DLL, we can omit this extra parameter entirely from the install_dll function.

cpp/Makefile
flibitijibibo: I dunno if it's GitHub's spacing or what but for some reason these two lines have two extra spaces on them, and I dunno if it's the site or the actual file...

flibitijibibo: Oh you know what it is, we have mixed tabs/spaces here, when the crap did that happen...

This one's not your fault, will deal with this later!

PR #61, opened 2018-11-12
Use libsamplerate for sample ratio conversion

aeikum:
This adds an optional dependency on [libsamplerate](http://www.mega-nerd.com/SRC/). This is a small, stable, high quality samplerate conversion library which is BSD licensed since 2016.

I'd highly recommend making this a non-optional dependency, so we can remove the linear resamplers and simplify the decoding/resampling code. I think it should be easy to port to your various platforms. If you're OK with that, I can begin that work (or you can do it).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,16 @@ ifeq ($(FAUDIO_FFMPEG), 1)
 	LDFLAGS += $(FFMPEG_LDFLAGS)
 endif
 
+# libsamplerate for high quality samplerate conversion
+ifeq ($(FAUDIO_LIBSAMPLERATE), 1)
+	LIBSAMPLERATE_CFLAGS = `pkg-config samplerate --cflags`
+	LIBSAMPLERATE_LDFLAGS = `pkg-config samplerate --libs`
+
+	LIBSAMPLERATE_CFLAGS += -DHAVE_LIBSAMPLERATE=1
+
+	CFLAGS += $(LIBSAMPLERATE_CFLAGS)
+	LDFLAGS += $(LIBSAMPLERATE_LDFLAGS)
+endif
 
 # Object code lists
 ifneq ($(FAUDIO_OUT),)
--- a/src/FAudio.c
+++ b/src/FAudio.c
@@ -196,6 +196,23 @@ void FAudio_UnregisterForCallbacks(
 	);
 }
 
+#if HAVE_LIBSAMPLERATE
+static uint32_t initialize_libsamplerate(FAudioSourceVoice *voice)
+{
+	int src_err;
+
+	voice->src.libsrc = src_new(SRC_SINC_FASTEST, voice->src.format->nChannels, &src_err);
+	if (!voice->src.libsrc)
+	{
+		return FAUDIO_E_UNSUPPORTED_FORMAT;
+	}
+
+	voice->src.resample = NULL;
+
+	return 0;
+}
+#endif
+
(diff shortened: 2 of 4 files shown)


aeikum:
Converting a 440Hz tone from 15000 samples per second to 44010 sps:

Linear resampler:

without



With libsamplerate:

with

flibitijibibo:
Holy hell, that linear resampler graph... I knew it was bad, but wow.

The only question I have is whether or not this is slower. AFAIK XAudio2 uses a linear resampler for speed, but if libsrc is just as fast then I'm okay with swapping it out. x86 may not notice much of a difference, but I bet ARM cares about this quite a bit. Even a few microseconds lost makes this nonviable.

In the case of the fixed-rate SRC this could actually _remove_ a dependency, as SDL2 dynamically loads this for SDL_AudioStream (and falls back if it fails). If we can use libsrc for both the dynamic _and_ fixed-rate converters, this removes the need for PlatformFixedRateSRC, which would make non-SDL2 support easier.

aeikum:
-24dB isn't really noticeable, but yeah, it adds up, especially with lots of sources or if you're using the source rate to simulate an engine revving up or something.

I'm sure it's slower than a linear conversion, but I didn't benchmark it. libsamplerate does provide several conversion options: http://www.mega-nerd.com/SRC/api_misc.html#Converters

See graphs below for the same conversion. Looks like SINC_FASTEST is probably our winner, assuming it's performant enough for you.

SINC_BEST
fa_best

SINC_MEDIUM
fa_src_med

SINC_FASTEST
fa_src_fast

LINEAR
fa_src_linear

FAudio's current linear
fa_lin

And just for fun, libsamplerate's ZOH:
fa_src_zoh

flibitijibibo:
Having the option list is a really big bonus too, I totally forgot about that! So maybe we can get fancy with the submix converters, and go with linear for the sources if we have to. If it's not fast enough, I suppose we could try to SIMD-ify the resampler (I'm surprised Epic didn't write something like that).

The only thing left on my list is how this affects the decoders. They're super sensitive (totally my fault) and we have to make sure we don't break the curBufferOffset stuff again. @JohanSmet found some issues with my recent FFmpeg stuff so I've asked them to check this out and see if we can't get yet another benefit out of this.

aeikum:
Force-pushed a rebase and a change to use FASTEST.

flibitijibibo:
Gave this a try and it seems to break things - current test is Dust: AET. When loading the game it freezes when attempting to do XACT work, and in some cases the audio becomes pure hissing. Might just need a quick Valgrind pass?

The latest Dust build on Steam should be compatible with upstream FAudio, so you can swap out the libFAudio files in lib64/lib depending on the architecture. You can also load the data in `Content/sfx/` with `facttool`.

flibitijibibo:
Good news on the performance front: When using `SRC_LINEAR` instead of `SRC_SINC_FASTEST`, the performance is pretty much the same on x86, as far as I can tell. I'm hoping ARM is just as good, but even naively-written NEON tends to make a pretty big difference for some reason.

When using SINC_FASTEST, performance took a _really_ big drop, so for sources that's definitely not happening. We can probably use it for submixes though.

aeikum:
Hm, rats. If we're stuck using linear resamplers for the sources, maybe it's best to just stick with our own code, since we can control its API.

flibitijibibo:
I think this ended up getting killed by performance... making our resampler/decoder process less sludgy would be nice though.

PR #63, opened 2018-11-12
check for iconv before copying or linking, fix bracket issue in wine_…

GloriousEggroll:
…setup_native

turns out iconv only gets used when compiling ffmpeg's avcodec if it's present, so we need to conditionally check if it exists before copying it over.

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -11,6 +11,9 @@ else
 	BUILD_DIR = build_win32
 endif
 
+#Mingw dir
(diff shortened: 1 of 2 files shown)


cpp/Makefile
```diff
@@ -11,6 +11,9 @@ else
 	BUILD_DIR = build_win32
 endif
 
+#Mingw dir
```
flibitijibibo: Missing a space here, the rest is fine!

PR #64, opened 2018-11-13
Debug logging take 2

aeikum:
A couple possibly controversial points. First we remove -pedantic to get rid of some useless compiler warnings. Next we create a new define, FAUDIO_RELEASE, to make compile-time decisions about optimizations. Finally this creates a global debug config struct so we don't need to dig up FAudio pointers in every function we want to write some logging from.

I skipped a lot of the granularity from the xaudio2 specification. I don't see most of that being useful in practice, and even MSDN says that it actually only supports the ERRORS mask. If we need more complicated functionality, that can be added later, I think.

I added a couple of useful debug messages around source voice creation and buffer consumption behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ endif
 ifeq ($(FAUDIO_RELEASE),1)
 	CFLAGS += -O3
 else
-	CFLAGS += -g -Wall -pedantic
+	CFLAGS += -g -Wall
 endif
 LDFLAGS += `sdl2-config --libs`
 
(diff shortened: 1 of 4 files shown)


flibitijibibo:
This at least gets us started; while 2.8+ logging is pretty useless, 2.7 produced some really interesting stuff that helped with FAudio's early development. That stuff can be added as needed (for example, @GloriousEggroll might need it for figuring out the Warframe buzzing).

PR #65, opened 2018-11-13
Disable __attribute__ on FAudio_debug_ for MSVC

0x0ade:
Simple fix when building with Visual C, which doesn't support `__attribute__`.

There's also (AFAIK) no drop-in replacement that'd mimic `format`.


0x0ade:
Just realized that I could've made the commit comment a little more descriptive...

flibitijibibo:
Might be worth ripping off SDL_stdinc.h at some point, but for now...

PR #66, opened 2018-11-14
[WIP] Initial XMA2 support

0x0ade:
**Note:** As long as the PR holds the WIP tag, it is not supposed to be merged.  
I'm already submitting it early to track progress and for general discussion.

This PR adds initial XMA2 support to FAudio using FFmpeg, similarly to how WMA support has been added by reusing most of the WMA support code. It also adds some required endian-swapping changes.

**Status:** XMA2 audio plays "fine" for the first few milliseconds (recorded to determine duration, but not yet compared against a functional conversion). It then then dies, repeating the last few samples endlessly.

Using "Death vs. The IRS" as a test game via XnaToFna and the latest version of FFmpeg:

image

My initial suspicion was that it's related to how FAudio handles the seek table. XnaToFna fed the seek table to FFmpeg via the `seek` chunk when converting the data ahead of time, while FAudio seems to perform all related seeking itself. I quickly dismissed that as being the cause, though, as the codepath in `FAudio_INTERNAL_DecodeFFMPEG` using `pDecodedPacketCumulativeBytes` is never being reached.

**CC:** @lamarqua


aeikum:
This might not be relevant, but some folks had trouble with my version of the ffmpeg patch in the old xaudio2 implementation. They "fixed" it by just ignoring errors, but afaict no one dug into why ffmpeg started spewing errors. https://bugs.winehq.org/show_bug.cgi?id=39402#c49

flibitijibibo:
I’m now reasonably convinced that #95 is the problem, not anything in this patch.

Maik, whenever you get a chance, go ahead and rebase and I’ll merge this in. If we can give them WMAv2/WMAv3/XMA all together that might help with getting support from the FFmpeg authors.

0x0ade:
My `xma` branch is now in sync again.

In semi-related news, it seems like FFmpeg randomly decided to die:

image

I'm having troubles replicating that, and it could be possible that it's been fixed in an FFmpeg update already - I haven't updated FFmpeg for quite some time.

PR #69, opened 2018-11-15
RFC: Add new entry point for clients to wrap the engine procedure

aeikum:
You can view the accompanying Wine changes at https://github.com/aeikum/wine/commits/engineproc

### DIFF
--- a/extensions/EngineProcedureEXT.txt
+++ b/extensions/EngineProcedureEXT.txt
@@ -1,0 +1,35 @@
+EngineProcedureEXT - Allow clients to wrap engine procedure
+
+About
+-----
+Clients may be interested in having more control over the engine procedure. In
+particular, they may want to modify or access the generated output before
+passing it to the underlying platform. In other cases, clients like Wine may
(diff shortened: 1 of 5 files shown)


extensions/EngineProcedureEXT.txt
```diff
@@ -0,0 +1,34 @@
+EngineProcedureEXT - Allow clients to wrap engine procedure
+
+About
+-----
+Clients may be interested in having more control over the engine procedure. In
+particular, they may want to modify or access the generated output before
+passing it to the underlying platform. Some clients, like Wine, may have
+thread requirements that SDL's threading API cannot satisfy. This extension
```
flibitijibibo: In addition to SDL not giving us enough control over the thread, certain platforms that generate their own audio thread also conflict with situations like Wine and BRUTE. For formality's sake, we can replace "SDL's threading" with "the platform's audio library".

Everything else looks good to me, just need `EXT` suffixes on the new types and functions.

aeikum:
OK, force pushed a new version with updated docs and symbol names.

PR #73, opened 2018-11-19
Add vim modelines

aeikum:
I understand if this is closed with "no, go away" :)


flibitijibibo:
Works On My vim, so...

PR #74, opened 2018-11-19
CMake modern target usage

NeroBurner:
Instead of variables use target_sources to add source files to an existing target.
Support multiple versions of SDL2 imported targets

Edit: Hope the spaced instead of tabs don't shock or anger you :see_no_evil:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,8 +19,8 @@ endif()
 # SDL2 Dependency
 find_package(SDL2)
 
-# Base FAudio Sources
-set(FAUDIO_SRC
+# Source lists
+add_library(FAudio SHARED
 	src/F3DAudio.c
 	src/FAudio.c
 	src/FAudio_internal.c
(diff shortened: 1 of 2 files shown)


flibitijibibo:
The tabs/spaces thing is fixable :P

The changes are fine but if they can be used by something first (utils folder?) then we'll know that the changes are meaningful too. See if this makes it any easier to build/run the utils; one thing I totally forgot about was the wavcommon paths:

https://github.com/FNA-XNA/FAudio/blob/master/utils/wavcommon/wavs.cpp#L4

NeroBurner:
updated, I can squash the commits if you'd like

I could have used globbing for the util executable, but I like explicit naming of files for cmake targets (also the header files, because then they show up in the IDE)

CMakeLists.txt
```diff
@@ -19,44 +19,60 @@ endif()
...
+elseif (TARGET SDL2)
+    message(STATUS "using TARGET SDL2")
+    target_link_libraries(FAudio PUBLIC SDL2)
+else()
+    message(STATUS "no TARGET SDL2::SDL2, or SDL2, using variables")
+    target_include_directories(FAudio PUBLIC ${SDL2_INCLUDE_DIRS})
+    target_link_libraries(FAudio PUBLIC ${SDL2_LIBRARIES})
+endif()
 
 # TODO: utils/
```
flibitijibibo: This should take care of the utils TODO. Are we able to folder the other CMakeLists.txt into this one? I try not to have nested CMake files unless the projects are actually separate from one another in a significant way.

Other than that, spaces->tabs should wrap this up!

NeroBurner:
for the wavs RESOURCE_PATH i'd recommend setting it by macro from cmake and only define it hard coded if it is not yet defined
```cpp
#ifndef RESOURCE_PATH
#ifdef _MSC_VER
#define RESOURCE_PATH "../../utils/wavcommon/resources"
#else
#define RESOURCE_PATH "utils/wavcommon/resources"
#endif
#endif 
```

flibitijibibo:
That'd work for me!

CMakeLists.txt
NeroBurner: sure, I don't like to have huge CMakeLists files, but it is your repo, so your wish is my command :)

NeroBurner: I remember why I like subfolder cmake lists files. I now have to add `utils/` before every source file

NeroBurner:
squashed and modified `wavs.cpp`

flibitijibibo:
Excellent, thanks for looking at this!

NeroBurner:
you're welcome :+1: 
Thanks for working on FAudio. Thanks to you (and all working on FAudio) I can play Warframe in 64bit mode ^_^

PR #75, opened 2018-11-19
cmake: add FFMPEG find script

NeroBurner:
add ffmpeg script and link to FAudio if needed


flibitijibibo:
Is the directory structure standard? Would it be okay if we changed it to cmakemodules/ to avoid the extra folder? Seems fine otherwise, though I wish this was a standard CMake module...

NeroBurner:
I wasn't able to find a written standard. LLVM is usig a subfolder for modules https://github.com/llvm-mirror/llvm/tree/master/cmake for example

flibitijibibo:
Very well... I'm hoping we don't need the added folder at all but the worst that'll happen is a git mv later.

PR #78, opened 2018-11-25
CMake install and config target

NeroBurner:
install FAudio shared library and generate a FAudio-config.cmake file
which can be used by other projects to link to FAudio library

commands to test:
```sh
cmake -H. -B_build -DCMAKE_INSTALL_PREFIX=${PWD}/_install
cmake --build _build --target install
```

generates an `_install` folder
```sh
tree _install
_install
├── include
│   └── FAudio.h
└── lib
    ├── cmake
    │   └── FAudio
    │       ├── FAudio-config.cmake
    │       ├── FAudio-targets.cmake
    │       └── FAudio-targets-noconfig.cmake
    └── libFAudio.so
```

edit: updated installed files tree with FAudio.h header file


flibitijibibo:
I'll have to defer to @aeikum on this one since I don't have experience with projects installing to the system and I want to make sure Proton can use this too... Andrew, does this look right to you?

aeikum:
I'll look at this closer later. I think if we're switching to cmake, we should remove all of the Makefiles at the same time, including the stuff for tests and utils and whatnot. I also noticed that your file listing is missing a bunch of header files that are installed with the current Makefile:
```
[aeikum@aeikum faudio64]$ find .
.
./lib
./lib/libFAudio.so
./include
./include/FAPOFX.h
./include/FACT.h
./include/FAPOBase.h
./include/FAudio.h
./include/FACT3D.h
./include/F3DAudio.h
./include/FAPO.h
./include/FAudioFX.h
```

NeroBurner:
Would it be a problem if we clean up a bit and put the header files into its own subfolder? In the repo a `FAudio` folder where all the public header files are in. And also when the files get installed a subfolder to avoid filename collitions. In the cmake config file we could add both paths (just `include` and `include/FAudio`) to the interface include folders for backwards compatibility

As for the missing header files, I'll add them later

flibitijibibo:
We can put them in `include/` if that makes things easier.

NeroBurner:
moved public headers and install them

also added header files to FAudio target, this informs the IDE to list the header files with the other source files
```
tree _install
_install
├── include
│   ├── F3DAudio.h
│   ├── FACT3D.h
│   ├── FACT.h
│   ├── FAPOBase.h
│   ├── FAPOFX.h
│   ├── FAPO.h
│   ├── FAudioFX.h
│   └── FAudio.h
└── lib
    ├── cmake
    │   └── FAudio
    │       ├── FAudio-config.cmake
    │       ├── FAudio-targets.cmake
    │       └── FAudio-targets-noconfig.cmake
    └── libFAudio.so
```

aeikum:
I made an Arch Linux PKGBUILD and the resulting package looked sane. I also integrated this branch into my Proton tree and it worked fine.

I would suggest removing the FORCE32 option and just have users use -DCMAKE_C_FLAGS="-m32" or similar.

I have an additional commit to build the host unit tests with cmake. I can push that after this is merged.

flibitijibibo:
That's good enough for me - @NeroBurner, if you can squish this down to two commits (one for the include folder and one for the other changes) I can merge this in.

NeroBurner:
@flibitijibibo done

PR #80, opened 2018-11-28
Always take into consideration PKG_CONFIG_FILE and use only pkg-confi…

GloriousEggroll:
…g to avoid unnecessary extra binary usage

In some cases I found PKG_CONFIG_FILE was not being taken into consideration (such as Arch's PKGBUILD system). This would cause pkg-config lines to detect incorrect libraries, even if PKG_CONFIG_PATH was specified. Example:

```
/usr/bin/ld: skipping incompatible /usr/lib/libSDL2.so when searching for -lSDL2
/usr/bin/ld: skipping incompatible /usr/lib/libavcodec.so when searching for -lavcodec
/usr/bin/ld: skipping incompatible /usr/lib/libavutil.so when searching for -lavutil
/usr/bin/ld: skipping incompatible /usr/lib/libgcc_s.so.1 when searching for libgcc_s.so.1
/usr/bin/ld: skipping incompatible /usr/lib/libpthread.so when searching for -lpthread
/usr/bin/ld: skipping incompatible /usr/lib/libpthread.a when searching for -lpthread
/usr/bin/ld: skipping incompatible /usr/lib/libc.so when searching for -lc
/usr/bin/ld: skipping incompatible /usr/lib/libc.a when searching for -lc
/usr/bin/ld: skipping incompatible /usr/lib/libgcc_s.so.1 when searching for libgcc_s.so.1
```
Further, if compiling with proton, using sdl2-config requires setting an extra environment bin path in addition to the extra lib path. We can avoid this by using `pkg-config sdl2` for our --libs and --cflags instead, just as other packages.

This should help various distro build systems by allowing them to specify PKG_CONFIG_PATH for 32 bit should the above error occur when compiling 32 bit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,12 @@ ifeq ($(FAUDIO_RELEASE),1)
 else
 	CFLAGS += -g -Wall
 endif
-LDFLAGS += `sdl2-config --libs`
+
+ifdef PKG_CONFIG_PATH
+	PKG_CONFIG_PATH ?= 
+endif
+
+LDFLAGS += `pkg-config sdl2 --libs`
 
 # Source lists
 FAUDIOSRC = \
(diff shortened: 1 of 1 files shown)


flibitijibibo:
I’m not 100% into this patch, mainly because sdl2-config should be properly installed first anyway...? The PKG_CONFIG_PATH check definitely feels hacky too. sdl2-config is, ironically, for improved portability, so I’m surprised it’s any different from pkg-config. Maybe we can check for custom pkg-config params and use that if they exist, otherwise use sdl2-config as what should be a safer default.

(Mind you, all of this will hopefully be made unnecessary with the CMake work... try out the stuff @NeroBurner has been working on and see if it’s better.)

flibitijibibo:
Very well - if the CMake stuff doesn’t move along quickly this week we can use this in the meantime. We should probably also update the Steam runtime to 2.0.9 if we can.

flibitijibibo:
I think the CMake stuff is moving along...

Issue #71, opened 2018-11-19
Replace Makefile with CMake


flibitijibibo:
I think enough cases have come up to consider replacing the Makefile... Typically I just write a handwritten Makefile because there's not much you're meant to do with the library, but FAudio has run into WAY more use cases than anything I've ever built before, so CMake is starting to make a little more sense now. There are many reasons for this, including easy Debug/Release configuration, a clear options list (FFmpeg, XNA_Song, etc), improving the utils/ builds, and if we can, integrating the cpp/ folder's builds.

Note that this is replacing ONLY the Makefiles. The VS/Xcode projects will stay as-is. (Maik, I'll still take the Android.mk if that's still useful)

I wrote up a basic file that covers the core sources, with a bunch of TODOs to fill in:

https://github.com/FNA-XNA/FAudio/tree/cmake

A bunch of people are involved, so here's a list so everyone has an idea of what's going on:
- Myself and @aeikum need this to work not just with Proton's build system, but with the needs of the various Linux distributions. If we're going to integrate FAudio into Wine, we need to make it easy for distros to package FAudio first. @GloriousEggroll has been working on the Proton build with FAudio's Makefile and has been hacking my build files to death, so we have a decent idea of what we need to fix to make that happen.
- @JohanSmet wrote the build system for the COM wrapper and Wine DLL; thankfully the two aren't wildly different but they do have some specific needs that have to be introduced just for those files somehow. @GloriousEggroll was playing with this a lot for Warframe, before we started looking at the Proton build.
- @cybik and @NeroBurner actually wrote their own CMake projects already, we just need to write a new CMakeLists.txt that covers the whole repo, not just the core FAudio sources.

As far as mandatory requirements go, we're stuck supporting Debian's criminally old CMake version, 3.7. Yes, it really is that old. Aside from this limitation, there aren't really any hard rules I can think of.

I'm currently stuck on the code side of things, but I know a lot of people were getting frustrated with my crappy old build system so I thought I'd open this up for anyone who's ready to throw it in the dumpster.

Checklist:
- [x] libFAudio
- [x] utils
- [x] tests
- [x] cpp (MinGW only)
- [x] FFmpeg

cybik:
> The VS/Xcode projects will stay as-is.

[...]

PR #86, opened 2018-12-06
 Replace Makefile with CMakeLists.txt

NeroBurner:
closes: https://github.com/FNA-XNA/FAudio/issues/71

Replace Makefile with CMakeLists.txt

@flibitijibibo please check my readme update and approve my cmake-generator flag commit. This generator flag also works with Visual Studio Builds, as those don't use `CMAKE_BUILD_TYPE`

To configure and build a Release build type with MSVC the following commands need to be used
```
cmake -H. -B_build_msvc -G "Visual Studio 15 2017 Win64"
cmake --build _build_msvc --config Release
```


flibitijibibo:
Ooh, glad VS is happy with this too - out of curiosity does this work with the cpp/ folder as well? I tried to clean it up to make that happen but I have no idea if I was actually successful.

In any case we have the premades to deal with this; long-term goal is to remove the premades but this was needed for Linux/macOS first and foremost.

Thanks again for all the work on this!

NeroBurner:
I did not actually try to build with MSVC, I just wanted to state how to build with MSVC and cmake. It took me a while to find those flags (especially the way how to actually BUILD from the command line)

Happy to spread the cmake-love :)

PR #93, opened 2018-12-11
[WIP] add CI using appveyor and travis-ci

NeroBurner:
Add Continuous Integration (CI) to automatically check every push (or pull request) if the code still compiles

This is a work in progress. Posting it for feedback

- [x] windows builds using appveyor
  - mingw-w64 amd64
  - msvc 2017 amd64
  - clang?
- [ ] linux builds using travis-ci
  - debian testing? We need something that has at least sdl 2.0.9
  - just for FAudio.so, or also for Utils?
- [ ] ~~upload artifacts of successful builds to github releases~~
  - [link for appveyor](https://www.appveyor.com/docs/deployment/github/)
- [ ] ~~add build status tags to README.md~~


flibitijibibo:
Much like SDL2# I'm up for hosting some CI files as long as someone else maintains it for their own purposes (build testing and whatnot). At the same time, GitHub releases will likely just be tags without binary releases, similar to FNA. I can keep fnalibs updated alongside each release, which is usually the 1st of every month.

For Linux/Windows builds, I use CentOS 7 which should have SDL 2.0.9 in EPEL.

flibitijibibo:
Here's the EPEL setup guide:

https://fedoraproject.org/wiki/EPEL

Then I believe it's `yum install git gcc SDL2-devel cmake3 mingw32-SDL2 mingw64-SDL2` to do a full build. You may have trouble with `mingwXX-cmake` since they don't have 3.x+, but thankfully I don't think we'll need to do MinGW builds going forward, we just need FAudio by itself. My hope is that the COM wrapper build will slowly get phased out as the Wine rewrite makes its way out (and hopefully into upstream).

flibitijibibo:
Let’s revisit this after we get FAudio in various distributons. My hope is that automated builds won’t be critical since everybody will just have FAudio alongside Wine, and thankfully FAudio is at a point where we aren’t needing a new build every day.

After distros have ripped up our build system, that’ll be a good time for third parties to add their own CI if they’d like. Like SDL2# I’ll host generic build files, just not the builds themselves.

PR #97, opened 2018-12-12
cpp: working dir independend setup script

NeroBurner:
assume the dlls are in the same folder as the script. This makes the
script work independent of the working directory

@GloriousEggroll please check confirm the changes

### DIFF
--- a/cpp/scripts/wine_setup_native
+++ b/cpp/scripts/wine_setup_native
@@ -10,8 +10,11 @@ if [ ! -f $WINEPREFIX/system.reg ]; then
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


cpp/scripts/wine_setup_native
```diff
@@ -10,8 +10,11 @@ if [ ! -f $WINEPREFIX/system.reg ]; then
 	exit 1
 fi
 
-winedump xaudio2_0.dll | grep 32BIT_MACHINE >> /dev/null 2>&1
+# assume the dlls are in the same folder as the executed script
```
flibitijibibo: I can see this breaking on my own setup - I do a build but not an install, so I'll be doing something like `../cpp/scripts/wine_setup_native` to avoid making extra copies of stuff.

PR #102, opened 2019-01-02
Rename README to README.md

vanillajonathan:


flibitijibibo:
Using plaintext over markdown was deliberate - it’s mostly a preference but I also don’t make any meaningful use of markdown (amd probably won’t without a very serious reason).

vanillajonathan:
The file arguably already is in Markdown format. If the file were to have .md file extension then GitHub would render it as such.

Salamek:
@flibitijibibo Hi, you are already using MARKDOWN format! :smile: 
Just rename that file to README.md and GitHub will render it correctly, you can test this on https://dillinger.io/, just paste raw README  into it to see the preview...

PR #122, opened 2019-03-25
Add missing NULL check in FAudioVoice_SetChannelVolumes

exolyte:
This fixes a crash in atelier firis: https://github.com/ValveSoftware/Proton/issues/1486

### DIFF
--- a/src/FAudio.c
+++ b/src/FAudio.c
@@ -1645,6 +1645,12 @@ uint32_t FAudioVoice_SetChannelVolumes(
 		LOG_API_EXIT(voice->audio)
 		return FAUDIO_E_INVALID_CALL;
 	}
+	
+	if (!voice->channelVolume)
+	{
+		LOG_API_EXIT(voice->audio)
+		return FAUDIO_E_INVALID_CALL;
+	}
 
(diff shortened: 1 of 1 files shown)


flibitijibibo:
This value shouldn't be NULL... the only reason it would be is if `outputChannels` is 0 or the call is being done on a mastering voice. I've dealt with the latter just now...

https://github.com/FNA-XNA/FAudio/commit/ca1729e416e27ffd15ecf35c6d69e8467db9cd29

... but I'm more concerned about the former. Here are the two spots where `outputChannels` is set, do either or these set 0, and if so when/why does it happen?

https://github.com/FNA-XNA/FAudio/blob/master/src/FAudio.c#L1195
https://github.com/FNA-XNA/FAudio/blob/master/src/FAudio.c#L1291

exolyte:
It looks like https://github.com/FNA-XNA/FAudio/commit/ca1729e416e27ffd15ecf35c6d69e8467db9cd29 fixes the issue, so I'll close the PR. Sorry for the noise.

flibitijibibo:
No sorrys needed, that lit up an assert that should have been real error handling. Thanks @exolyte!

PR #123, opened 2019-03-26
mingw: wine_setup_native this should just be iconv.dll

GloriousEggroll:


flibitijibibo:
Can we have both and not just one? This is totally going to break another build setup.

PR #131, opened 2019-04-16
dump audio buffer to wav files

NeroBurner:
- use a single function to open the same file for the same format
- write extensible data if file format is extended
- open file once for each submitChunk
- use only SDL_RWops functions for file access

It's not creating working files. It creates readable headers for non-extended encodings (at least VLC recognizes the format, but is not able to load the files). Warframe uses ADPCM and WMAUDIO2. I unfortunately havn't found refernce files for these formats to trim the dumping to produce working wma files

Feedback and tips are very welcome

### DIFF
--- a/src/FAudio.c
+++ b/src/FAudio.c
@@ -52,6 +52,18 @@ MAKE_SUBFORMAT_GUID(WMAUDIO3, FAUDIO_FORMAT_WMAUDIO3);
 MAKE_SUBFORMAT_GUID(WMAUDIO_LOSSLESS, FAUDIO_FORMAT_WMAUDIO_LOSSLESS);
 #undef MAKE_SUBFORMAT_GUID
 
+#ifdef FAUDIO_DUMP_VOICES
+static void FAudio_DUMPVOICE_Init(const FAudioSourceVoice *voice);
+static void FAudio_DUMPVOICE_Finalize(const FAudioSourceVoice *voice);
+static void FAudio_DUMPVOICE_WriteBuffer(
+	const FAudioSourceVoice *voice,
+	const FAudioBuffer *pBuffer,
+	const FAudioBufferWMA *pBufferWMA,
+	const uint32_t playBegin,
+	const uint32_t playLength
+);
+#endif /* FAUDIO_DUMP_VOICES */
+
 /* FAudio Version */
 
 uint32_t FAudioLinkedVersion(void)
@@ -509,6 +521,10 @@ uint32_t FAudio_CreateSourceVoice(
 		audio->pMalloc
 	);
 	FAudio_AddRef(audio);
+
+#ifdef FAUDIO_DUMP_VOICES
+	FAudio_DUMPVOICE_Init(*ppSourceVoice);
+#endif /* FAUDIO_DUMP_VOICES */
 
 	LOG_API_EXIT(audio)
 	return 0;
@@ -2049,6 +2065,10 @@ void FAudioVoice_DestroyVoice(FAudioVoice *voice)
 	{
 		FAudioBufferEntry *entry, *next;
 
+#ifdef FAUDIO_DUMP_VOICES
+		FAudio_DUMPVOICE_Finalize((FAudioSourceVoice*) voice);
+#endif /* FAUDIO_DUMP_VOICES */
+
 		FAudio_PlatformLockMutex(voice->audio->sourceLock);
 		LOG_MUTEX_LOCK(voice->audio, voice->audio->sourceLock)
 		while (voice == voice->audio->processingSource)
@@ -2392,6 +2412,14 @@ uint32_t FAudioSourceVoice_SubmitSourceBuffer(
 		entry->buffer.LoopCount = 0;
 	}
 
+#ifdef FAUDIO_DUMP_VOICES
+	/* dumping current buffer, append into "data" section */
+	if (pBuffer->pAudioData != NULL && playLength > 0)
+	{
+		FAudio_DUMPVOICE_WriteBuffer(voice, pBuffer, pBufferWMA, playBegin, playLength);
+	}
+#endif /* FAUDIO_DUMP_VOICES */
+
 	/* Submit! */
 	FAudio_PlatformLockMutex(voice->src.bufferLock);
 	LOG_MUTEX_LOCK(voice->audio, voice->src.bufferLock)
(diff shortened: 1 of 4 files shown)


flibitijibibo:
Other than it not working, three major things:

1: This needs to be guarded with a DUMP_VOICES build option
2: The SDL dependency needs to be removed, FAudioIOStream should be usable for this
3. General style fixes and C compliance issues, try to match the rest of the source and check against various compilers with strict rules (VS2010 is my go-to since it has super basic C support)

Without those three things and without working files, this isn’t too useful and it’ll just break all of our builds.

NeroBurner:
Small update:

1. added DUMP_VOICES build option
2. moved the SDL dependency into FAudio_platform_sdl2.c, but I needed to create a FAudioIOStreamOut wrapper. I dared not touch the FAudioIOStream struct in the FAudio.h header
3. still missing, I redid the comments in the /**/ format, still need to clean up the comments a bit more, but no pedantic warnings on my changes

I can now successfully write and play voices from Warframe dumped through FAudio (Format 0x0002 adpcm_ms is working, WMAv2 plays, but is just static, I suspect I still need to write the "dpds" structure and dump the bufferWMA struct into it)

```
ffprobe FA_fmt_0x0002_0x0000_0x0000000005C1BA60.wav ffprobe version n4.1.3 Copyright (c) 2007-2019 the FFmpeg developers
  built with gcc 8.2.1 (GCC) 20181127
  configuration: --prefix=/usr --disable-debug --disable-static --disable-stripping --enable-fontconfig --enable-gmp --enable-gnutls --enable-gpl --enable-ladspa --enable-libaom --enable-libass --enable-libbluray --enable-libdrm --enable-libfreetype --enable-libfribidi --enable-libgsm --enable-libiec61883 --enable-libjack --enable-libmodplug --enable-libmp3lame --enable-libopencore_amrnb --enable-libopencore_amrwb --enable-libopenjpeg --enable-libopus --enable-libpulse --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libv4l2 --enable-libvidstab --enable-libvorbis --enable-libvpx --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxcb --enable-libxml2 --enable-libxvid --enable-nvdec --enable-nvenc --enable-omx --enable-shared --enable-version3
  libavutil      56. 22.100 / 56. 22.100
  libavcodec     58. 35.100 / 58. 35.100
  libavformat    58. 20.100 / 58. 20.100
  libavdevice    58.  5.100 / 58.  5.100
  libavfilter     7. 40.101 /  7. 40.101
  libswscale      5.  3.100 /  5.  3.100
  libswresample   3.  3.100 /  3.  3.100
  libpostproc    55.  3.100 / 55.  3.100
Input #0, wav, from 'FA_fmt_0x0002_0x0000_0x0000000005C1BA60.wav':
  Duration: 00:00:05.19, bitrate: 318 kb/s
    Stream #0:0: Audio: adpcm_ms ([2][0][0][0] / 0x0002), 39748 Hz, 2 channels, s16, 317 kb/s
```

and the shortened WMAv2 file
```
ffprobe FA_fmt_0xFFFE_0x0161_0x0000000005C1C2B0.wav
[wav @ 0x55651e995640] Estimating duration from bitrate, this may be inaccurate
Input #0, wav, from 'FA_fmt_0xFFFE_0x0161_0x0000000005C1C2B0.wav':
  Duration: 00:00:23.41, bitrate: 192 kb/s
    Stream #0:0: Audio: wmav2 (a[1][0][0] / 0x0161), 44100 Hz, 2 channels, fltp, 192 kb/s
```

src/FAudio.c
```diff
@@ -240,6 +240,185 @@ void FAudio_UnregisterForCallbacks(
...
+		const FAudioWaveFormatExtensible *format_ex =
+				(const FAudioWaveFormatExtensible*) format;
+		format_ex_tag = (uint16_t) (format_ex->SubFormat.Data1);
+	}
+	if (format->wFormatTag == FAUDIO_FORMAT_WMAUDIO2)
+	{
+		format_tag = FAUDIO_FORMAT_EXTENSIBLE;
+		format_ex_tag = FAUDIO_FORMAT_WMAUDIO2;
+	}
+	SDL_snprintf(
```
flibitijibibo: `FAudio_snprintf` is available to use in place of direct SDL_snprintf.

flibitijibibo:
This is definitely an improvement! Notes for review pass 2:

- DUMP_VOICES guard should be more aggressive, basically encompassing all new code
- We may have to put all the DUMP_VOICES stuff at the bottom, with forward declarations right above the source voice functions, just to make life easier for day-to-day development
- Yep, dpds should be possible to generate with the bufferWMA contents, though it'll be fun figuring out how to append that in between the header and data...

NeroBurner:
- remove SDL_snprintf
- more aggressive `DUMP_VOICES`
- forward declaration and DumpVoices functions at bottom of `FAudio.c` file
- write BufferWMA to dpds section (done by writing the dpds and data section into its own files and in the finalize step append those files to the result wav file)

the WMAv2 files still only contain gibberish noises, but the uncompressed file formats are (nearly) perfect. I'm out of ideas on the WMAv2 front. Any help would be highly appreciated

src/FAudio.c
NeroBurner: replaced, good catch

flibitijibibo:
The only idea I really have is to try loading a RIFF WMA file from here...

https://github.com/FNA-XNA/FAudio/blob/master/utils/testxwma/testxwma.cpp

... then do a bindiff of the original file compared to the source output.

NeroBurner:
tested with a MSADPCM (format 0x0002) file. The dumped file still plays and is nearly identical. The header isn't completely the same and the file length of the dumped file is shorter. Therefore my question

in case of MSADPCM the playlength is computed here
https://github.com/FNA-XNA/FAudio/blob/de7687aee5baea3c0ef5d0fa29048660aa4201dd/src/FAudio.c#L2277-L2285

and later the playlength (and other play and loop variables) are aligned here
https://github.com/FNA-XNA/FAudio/blob/de7687aee5baea3c0ef5d0fa29048660aa4201dd/src/FAudio.c#L2329-L2340

Could it be that we cap the playlength two times? Maybe we can remove the special handling for playlength (the first source snippet) and just keep the aligning calculation?

NeroBurner:
@flibitijibibo please merge it already worked, but VLC does not like to play the extended WMA2 files `FA_fmt_0xFFFE_0x0161`

I was able to dump `FA_fmt_0xFFFE_0x0161` files and then successfully play them with `testxwma`

PR #134, opened 2019-05-01
[WIP] Implemented OperationSets/CommitChanges

TylerGlaiel:
implemented operationSet stuff


TylerGlaiel:
I have not tested this thoroughly, don't merge it before you do!

flibitijibibo:
TODO List:

- [ ] Figure out what the Wine team wants me to do with all my 19.06 breakages
- [ ] Add new files to projects
- [ ] Test for possible deadlocks (some might be accurate to XAudio2, I'll probably just test with Puyo Puyo Tetris again)
- [ ] Tabbing, pedantic C compliance, blah blah blah

TylerGlaiel:
oh yeah this probably breaks if you call commit changes from a different thread than you're queueing commands from

flibitijibibo:
Looking at some other XAudio2 applications, it appears operation parameters are copied when building the command. For example, SetOutputMatrix copies the matrix array instead of storing just the pointer:

https://github.com/id-Software/DOOM-3-BFG/blob/1caba1979589971b5ed44e315d9ead30b278d8b4/neo/sound/XAudio2/XA2_SoundVoice.cpp#L300

So we’ll have to allocate a little bit more when creating commands with dynamically-sized parameters.

TylerGlaiel:
ok that's not every parameter however, SetEffectParameters explicitly says that the pointer must remain valid https://docs.microsoft.com/en-us/windows/desktop/api/xaudio2/nf-xaudio2-ixaudio2voice-seteffectparameters

the ones that look like they would need to copy are:
SetFilterParameters (one struct, not variable size, probably does not need an additional allocation)
SetChannelVolumes (needs an allocation)
SetOutputMatrix (needs an allocation)

im just gonna do it as a second allocation for both of those 2 instead of doing variable length struct shenanigans

flibitijibibo:
The remark for that function is, as far as I can tell, not actually true - they appear to use a staging buffer to actually pass the data to the XAPO, and it looks like that buffer is filled at call time, regardless of the operation set. For example...

```
float derp = 1.0f;
voice->SetEffectParameters(0, &derp, sizeof(derp), 1);
derp = 69.0f;
engine->CommitChanges(1);
```

IXAPOParameters::SetParameters will still get 1.0f instead of 69.0f from a totally unique pointer, and the effect won't get the new value until you call SetEffectParameters again. In fact, I can even use a temporary value like the above in a small function and it still works without getting corrupted.

flibitijibibo:
First review is done, should cleanly apply on top of your fork:

https://github.com/TylerGlaiel/FAudio/pull/1

Codewise this now looks good to me, so now we get to test. I'll probably do this throughout the next week, and if everything still works I'll merge on May 10.

flibitijibibo:
Oh also CC @TheSpydog since this affects the Xcode projects

TylerGlaiel:
fixed something that didn't survive the style update

things to test: 
queuing / committing operations from within a callback (the operationLock mutex is locked during callbacks, I think)
make sure it actually can start sounds atomically (do like sleep(500ms) after every operation in Commit) (otherwise the platform needs to be locked as well)

flibitijibibo:
The latest changes fix the obvious contention issue we had before, so as long as the style's all good and we can't come up with new ways to deadlock (callbacks within the execute function maybe?) this should be good to go.

TylerGlaiel:
one thing- with the new changes all the operations that use OperationSet happen asynchronously, so there's an issue when you destroy a voice that it could still have pending unexecuted operations in either queue (even if you made sure to commit them beforehand), as all the operations have FAudioVoice* as their first parameter, you should probably go through and remove all the ones relevant to a voice when it gets destroyed

flibitijibibo:
This is true - should we just add a ClearAllForVoice and pull the FAudioVoice* pointer as a shared value in the struct? (The SourceVoice cases aren't worrisome, since they're just the same type with different names.)

TylerGlaiel:
you could though just accessing it through one of the unions in the struct is probably fine and would result in less changes needed to the code. There's 2 that have a second voice as a parameter that would need to be checked separately (SetOutputVoices, SetOutputMatrix)

flibitijibibo:
Pushing this to next week - still have to write ClearAllForVoice, style pass 2, and one last QA pass.

flibitijibibo:
Latest push takes care of the ClearAll and the last of the style stuff. Going to spend _all_ of tomorrow on testing, and provided I don't see anything in 8 straight hours of desperately trying to get this to deadlock, I'll merge on Friday.

flibitijibibo:
Everything seems to work now. Upon closer review it seems like there shouldn't be any new deadlocks thanks to the Execute function, and any deadlock that might occur would actually be the fault of other FAudio components and not the operation set work (particularly since no operation has callbacks internally), so it's not really this patch's concern.

Thanks for looking into this @TylerGlaiel!

PR #139, opened 2019-06-09
one more extraneous semicolon

rfht:
found one more...


rfht:
that was quick for the world's tiniest PR... :]

PR #140, opened 2019-06-12
faudio: actually trust cbsize

NeroBurner:
is there a reason we don't trust the user to provide a valid cbSize variable?


flibitijibibo:
Does this fix anything in particular?

flibitijibibo:
Actually I remember this, this is our own format declaration so we should be using our own size. For extensible formats provided by the client we do copy their cbSize.

NeroBurner:
no actual bug, just stumbled upon it while dumping the buffers into RIFF WAVE files and creating the WAVEFORMATEX/WAVEFORMATEXTENSIBLE headers. Thanks for the clarification

PR #142, opened 2019-06-14
remove not needed semicolon after LOG functions

NeroBurner:


src/FAudio.c
```diff
@@ -2330,9 +2330,9 @@ uint32_t FAudioSourceVoice_SubmitSourceBuffer(
 	LOG_INFO(
 		voice->audio,
 		"%p: appended buffer %p",
-		voice,
-		&entry->buffer
-	);
+		(void *) voice,
```
flibitijibibo: Extra typecasts all of a sudden...? Other than that this patch lgtm.

NeroBurner: this is a `-Wformat` warning

from another place, but the same warning (just for reference):
```
FAudio_internal.h:571: warning: format ‘%p’ expects argument of type ‘void *’, but argument 7 has type ‘FAudioVoice *’ {aka ‘struct FAudioVoice *’} [-Wformat=]
 #define LOG_ERROR(engine, fmt, ...) PRINT_DEBUG(engine, ERRORS, "ERROR", fmt, __VA_ARGS__)
                                                                 ^~~~~~~
```

will move the casts to the other PR

NeroBurner: moved it to #143

PR #143, opened 2019-06-14
fix conversion warning

NeroBurner:
- explicit cast from uint32_t to doube of parameter
- fix warning of function cast by putting brackets around the command

### DIFF
--- a/src/FAudio.c
+++ b/src/FAudio.c
@@ -456,12 +456,12 @@ uint32_t FAudio_CreateSourceVoice(
 	}
 
 	/* Sample Storage */
-	(*ppSourceVoice)->src.decodeSamples = (uint32_t) FAudio_ceil(
-		audio->updateSize *
+	(*ppSourceVoice)->src.decodeSamples = (uint32_t) (FAudio_ceil(
+		(double) audio->updateSize *
 		(double) MaxFrequencyRatio *
 		(double) (*ppSourceVoice)->src.format->nSamplesPerSec /
 		(double) audio->master->master.inputSampleRate
-	) + EXTRA_DECODE_PADDING * (*ppSourceVoice)->src.format->nChannels;
+	)) + EXTRA_DECODE_PADDING * (*ppSourceVoice)->src.format->nChannels;
 	FAudio_INTERNAL_ResizeDecodeCache(
 		audio,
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Which compiler is emitting this stuff, anyway...?

If you want, you can throw the (void*) casts from #142 into this patch instead. But I'm curious to know what compiler is complaining, I'm not seeing this with VS2010/GCC/Clang.

src/FAudio.c
```diff
@@ -456,12 +456,12 @@ uint32_t FAudio_CreateSourceVoice(
...
 	/* Sample Storage */
-	(*ppSourceVoice)->src.decodeSamples = (uint32_t) FAudio_ceil(
-		audio->updateSize *
+	(*ppSourceVoice)->src.decodeSamples = (uint32_t) (FAudio_ceil(
+		(double) audio->updateSize *
 		(double) MaxFrequencyRatio *
 		(double) (*ppSourceVoice)->src.format->nSamplesPerSec /
 		(double) audio->master->master.inputSampleRate
-	) + EXTRA_DECODE_PADDING * (*ppSourceVoice)->src.format->nChannels;
+	) + EXTRA_DECODE_PADDING * (*ppSourceVoice)->src.format->nChannels);
```
flibitijibibo: Oh also! This added paren should be at the end of FAudio_ceil() and not the whole statement, since we're trying to round up for fractions of a sample and _then_ do addition/multiplication on top of that safety padding.

NeroBurner: fixed

flibitijibibo:
Not sure why but it says this can’t be merged cleanly... after a rebase this should be good to go, I think.

NeroBurner:
the warnings come from clang (and most probably would also come from gcc). I'm using qtcreator and it has a relatively new "Clang Code Model" which by default uses clang with the following options

```
-Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-unused-macros -Wno-newline-eof -Wno-exit-time-destructors -Wno-global-constructors -Wno-gnu-zero-variadic-macro-arguments -Wno-documentation -Wno-shadow -Wno-switch-enum -Wno-missing-prototypes -Wno-used-but-marked-unused
```

NeroBurner:
rebased

PR #145, opened 2019-06-14
Fix warnings

NeroBurner:
found some warnings, kept them in small commits to easier rebase an squash them (or remove if one isn't up to snuff)

### DIFF
--- a/src/FAudio_internal.c
+++ b/src/FAudio_internal.c
@@ -411,8 +411,8 @@ static void FAudio_INTERNAL_DecodeBuffers(
 		LOG_INFO(
 			voice->audio,
 			"Voice %p, buffer %p, decoded %u samples from [%u,%u)",
-			voice,
-			buffer,
+			(void *) voice,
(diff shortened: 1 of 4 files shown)


src/FAudio_internal.c
```diff
@@ -90,7 +90,7 @@ void FAudio_INTERNAL_debug(
 	va_end(va);
 
 	/* Print, finally. */
-	FAudio_Log(output);
+	FAudio_Log(output)
```
flibitijibibo: This is an error, isn't it?

flibitijibibo: Oh hey, turns out it isn't, it's a bug in my macros:

https://github.com/FNA-XNA/FAudio/blob/master/src/FAudio_internal.h#L79
https://github.com/FNA-XNA/FAudio/blob/master/src/FAudio_internal.h#L140

NeroBurner: Happy to help find bugs :)

Issue #144, opened 2019-06-14
Minimum C standard C89, C99, C11


NeroBurner:
What's the required minimum C-standard?

Found the following warning when enabling the -pedantic switch:
```
FAudio_internal.h:482: warning: ISO C99 doesn’t support unnamed structs/unions [-Wpedantic]
  };
   ^
```

anonymous unions are a C11 extension

flibitijibibo:
It's supposed to be C99, but naming that union would be... quite the addition. I'd like to silence this one for Wpedantic if possible, since all supported compilers are okay with this.

NeroBurner:
I'm trying to follow your feedback on the DumpVoices PR

> 3. General style fixes and C compliance issues, try to match the rest of the source and check against various compilers with strict rules (VS2010 is my go-to since it has super basic C support)

But I don't want to install VS2010 to test compliance. So I tried to add warning flags

NeroBurner:
couldn't we just push the requirement to C11?

flibitijibibo:
You can for your qtcreator project if you want, but honestly if the only thing pushing us to C11 is that the compiler doesn't like `union { };` I'm not really sure I care about Wpedantic in this case. If you really want, you can go figure out a way to incorporate something like `DUMMYUNIONNAME` and `NONAMELESSUNION` like Wine does, but even this will be a lot more work than you think it'll be.

flibitijibibo:
Looked into this some more, would simply changing C_STANDARD to gnu99 resolve this?

NeroBurner:
does MSVC support gnu extensions?

and I think it is not enough
```
/bin/cc -DFAUDIO_DISABLE_DEBUGCONFIGURATION -DFAudio_EXPORTS -DHAVE_FFMPEG=1 -I/home/nero/repos/FAudio/src -I/home/nero/repos/FAudio/include -I/home/nero/repos/FAudio/SYSTEM -isystem /usr/include/SDL2  -O3 -DNDEBUG -fPIC   -Wall -Wextra -pedantic -Wno-unused-parameter -Wno-unused-but-set-variable -std=gnu99 -o CMakeFiles/FAudio.dir/src/XNA_Song.c.o   -c /home/nero/repos/FAudio/src/XNA_Song.c
In file included from /home/nero/repos/FAudio/src/XNA_Song.c:29:
[...]

PR #146, opened 2019-06-16
silence pedantic warnings for unnamed union and structs

NeroBurner:
using __extension__ keyword
Found solution at: https://stackoverflow.com/a/20345704

https://github.com/FNA-XNA/FAudio/issues/144

### DIFF
--- a/include/FACT.h
+++ b/include/FACT.h
@@ -64,9 +64,9 @@ typedef struct FACTOverlapped
 {
 	void *Internal; /* ULONG_PTR */
 	void *InternalHigh; /* ULONG_PTR */
-	union
+	__extension__ union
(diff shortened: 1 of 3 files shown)


flibitijibibo:
This will break on MSVC, if not Clang in some scenarios as well. You will most likely have to create something like

```
#if defined(SOME CHECK FOR GCC HERE)
#define FAUDIONAMELESS __extension__
#else
#define FAUDIONAMELESS
#endif
```

and use that instead.

NeroBurner:
added it

flibitijibibo:
I’ll probably pull this into one def in FAudio.h, but as long as we have this...!

PR #147, opened 2019-06-16
Fix stb.h warnings

NeroBurner:
fix signed/unsigned comparison warning in stb.h

also move the defines into stb.h header itself (makes the clang code model happy and stb.h can be used without need to define stuff)

### DIFF
--- a/src/FACT_internal.c
+++ b/src/FACT_internal.c
@@ -29,8 +29,6 @@
 
 /* RNG */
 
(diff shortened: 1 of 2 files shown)


flibitijibibo:
It's our policy not to break compatibility with upstream stb.h unless we absolutely have to - if the unsignedness is an issue, take it up with Sean: https://github.com/nothings/stb

PR #149, opened 2019-06-16
internal: use unsigned for index variable

NeroBurner:
use an unsigned integer for index variable
and fix -Wsign-compare warning

### DIFF
--- a/src/FACT_internal.c
+++ b/src/FACT_internal.c
@@ -165,7 +165,7 @@ void FACT_INTERNAL_GetNextWave(
 	uint8_t loopCount = 0;
 	float max, next;
 	uint8_t noTrackVariation = 1;
-	int32_t i;
+	uint32_t i;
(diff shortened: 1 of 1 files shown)


flibitijibibo:
If the output asm is okay with this then I'll merge, otherwise we'll need to come up with a way to make this not slow (which at a glance this looks like it will be).

NeroBurner:
the compiler seems to be smart enough to handle the [i-1] calls. It seems it even results in a small performance increase (with `-O3`)

http://quick-bench.com/lUtgp-Q3MpNtLkQjZ9si7Zvs6IA

https://godbolt.org/z/GZ_APG

flibitijibibo:
I still worry how maintainable this is, but it does look like the difference is negligible even in the real source (at least from what I can see, VS2010 probably sucks at this).

Issue #153, opened 2019-07-11
SDL2_INCLUDE_DIRS and SDL_LIBRARIES should override sdl2-config


madewokherd:
We use SDL2_INCLUDE_DIRS and SDL2_LIBRARIES in wine-mono to supply our own SDL2: https://github.com/madewokherd/wine-mono/blob/master/Makefile#L177

We use a similar pattern in CrossOver, except that we don't cross-compile. If a 64-bit SDL2 is available but a 32-bit SDL2 is not, FAudio tries to use the system SDL2 for the 32-bit build, which doesn't work.

The system SDL2 should be ignored in all cases when these variables are specified anyway, for consistency.

flibitijibibo:
We most likely just have to refactor this to go a totally different path when the variables are explicitly set: https://github.com/FNA-XNA/FAudio/blob/master/CMakeLists.txt#L158

Not too hard, but we’ll have to make sure it doesn’t break on the known setups.

flibitijibibo:
Threw this together:

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index d3c130a..037adce 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,28 +156,35 @@ if(FFMPEG)
 endif(FFMPEG)
 
 # SDL2 Dependency
-if (NOT DEFINED SDL2_INCLUDE_DIRS AND NOT DEFINED SDL2_LIBRARIES)
-	find_package(SDL2 CONFIG)
-endif()
-if (TARGET SDL2::SDL2)
-	message(STATUS "using TARGET SDL2::SDL2")
-	target_link_libraries(FAudio PUBLIC SDL2::SDL2)
-	if(INSTALL_MINGW_DEPENDENCIES)
-		install_shared_libs(bin SDL2::SDL2)
-	endif()
-elseif (TARGET SDL2)
-	message(STATUS "using TARGET SDL2")
-	target_link_libraries(FAudio PUBLIC SDL2)
-	if(INSTALL_MINGW_DEPENDENCIES)
-		install_shared_libs(bin SDL2)
-	endif()
-else()
-	message(STATUS "no TARGET SDL2::SDL2, or SDL2, using variables")
+if (DEFINED SDL2_INCLUDE_DIRS AND DEFINED SDL2_LIBRARIES)
 	target_include_directories(FAudio PUBLIC "$<BUILD_INTERFACE:${SDL2_INCLUDE_DIRS}>")
[...]

PR #154, opened 2019-07-16
improved install shared libs

NeroBurner:
using the cmake attention from https://github.com/FNA-XNA/FAudio/issues/153 to promote an improvement on the `install_shared_libs` function. It's now more verbose, more features and more like the cmake `install()` command

this branch is based on `cmakefix`


flibitijibibo:
Totally fell asleep on this, sorry...

@madewokherd, does this work for you? If so I can merge; most of this is just for the COM wrapper but as long as the wine-mono build is affected this should be fine.

flibitijibibo:
This works on my build environments and worked for the test case that was reported, so in this goes...

PR #162, opened 2019-12-12
testxwma: fix buffer overflow

NeroBurner:
When a file with a fmt chunkSize > 40 (size of FAudioWaveFormatExtensible struct)
was processed a buffer overflow occured potentially corrupting memory.


flibitijibibo:
Is wfx freed at some point?

NeroBurner:
now it is :sweat_smile:

PR #165, opened 2020-01-08
XACT support for xactengine 3.1

alesliehughes:
When investigating the following wine issue
https://bugs.winehq.org/show_bug.cgi?id=38615

It's attempting to load older XACT files which is causing the demo to crash on startup.
The data is the size, so I'm assuming that extra flags are available for the newer XACT format.

I've also changed the return values from -1 to -4 for each return branch, so we can workout why a function failed from the wine side.

### DIFF
--- a/src/FACT_internal.c
+++ b/src/FACT_internal.c
@@ -2940,8 +2940,8 @@ uint32_t FACT_INTERNAL_ParseWaveBank(
 		}
 	}
 	if (	header.dwSignature != 0x444E4257 ||
-		header.dwVersion != FACT_CONTENT_VERSION ||
-		header.dwHeaderVersion != 44	)
+		(header.dwVersion != FACT_CONTENT_VERSION && header.dwVersion != FACT_CONTENT_VERSION_3_1) ||
+		(header.dwHeaderVersion != 44 && header.dwHeaderVersion != 42) )
 	{
 		return -1; /* TODO: NOT XACT FILE */
 	}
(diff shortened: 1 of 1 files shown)


src/FACT_internal.c
```diff
@@ -1963,13 +1963,24 @@ uint32_t FACT_INTERNAL_ParseAudioEngine(
...
-	if (	(magic != 0x46534758 && !(se = (magic == 0x58475346))) /* 'XGSF' */ ||
-		read_u16(&ptr, se) != FACT_CONTENT_VERSION ||
-		read_u16(&ptr, se) != 42 /* Tool version */	)
+	if (magic != 0x46534758)
 	{
 		return -1; /* TODO: NOT XACT FILE */
 	}
 
+	content = read_u16(&ptr, se);
+	if (content != FACT_CONTENT_VERSION &&
```
flibitijibibo: This line’s `content` can be tabbed too

src/FACT_internal.c
```diff
@@ -2399,19 +2410,31 @@ uint32_t FACT_INTERNAL_ParseSoundBank(
...
-		read_u16(&ptr, se) != 43 /* Tool version */	)
+
+	if (magic != 0x4B424453)
 	{
-		return -1; /* TODO: NOT XACT FILE */
+            return -1;
+	}
+
+	content = read_u16(&ptr, se);
+	if (content != FACT_CONTENT_VERSION &&
```
flibitijibibo: This line’s `content` can be tabbed too

flibitijibibo:
Excellent, thanks!

PR #169, opened 2020-01-13
Wavebank shouldn't destroy the parentCue as it's content may already have been destroyed

alesliehughes:
When FACT_INTERNAL_CreateSound creates a Wave, it assigns the Cue to wave->parentCue.  If you then call FACT_CueDestroy on the Cue, assume a Play/Stop/Destroy. Later when the WaveBank is destroyed, it loops the waves and checks wave->parentCue, however this is no longer valid and causes a crash when its attempts to destroy it again.

There might be better way to clear wave->parentCue when FACT_CueDestroy is called, happy for comments.

### DIFF
--- a/src/FACT.c
+++ b/src/FACT.c
@@ -1297,15 +1297,7 @@ uint32_t FACTWaveBank_Destroy(FACTWaveBank *pWaveBank)
 	while (pWaveBank->waveList != NULL)
 	{
 		wave = (FACTWave*) pWaveBank->waveList->entry;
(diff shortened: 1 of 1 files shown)


flibitijibibo:
When the Cue is destroyed it should destroy the wave, which should then be cleared from the list. It sounds like there’s a leak in this case, this patch would only cause corruption when clearing Cues instead.

alesliehughes:
Thanks.  Ill have another look at it.

PR #172, opened 2020-01-15
FACTAudioEngine_RegisterNotification handle FACT_FLAG_NOTIFICATION_PERSIST

alesliehughes:
We need to variable to handle each of FACT_FLAG_NOTIFICATION_PERSIST types, but for now it good enough to not attempt to access members that possible invalid.

Would it be OK to put in the engine 
uint32 notication_flags  -- 0x01, 0x02, etc and
void * wb_context
variable for each type?


flibitijibibo:
This patch looks good to me. I'm not sure I understand the question though - send it as a separate patch and I can check it out.

PR #179, opened 2020-02-19
Allow for building a static library instead of shared

hcorion:
Over at https://github.com/RPCS3/rpcs3 we use FAudio in a submodule, but we only want to link statically.
We don't want to use the shared library because when you `make install` rpcs3 will conflict with Linux FAudio packages.
This PR keeps the default functionality of building shared libraries, but just adds the ability to build FAudio statically as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ option(XNASONG "Build with XNA_Song.c" ON)
 option(LOG_ASSERTIONS "Bind FAudio_assert to log, instead of platform's assert" OFF)
 option(FORCE_ENABLE_DEBUGCONFIGURATION "Enable DebugConfiguration in all build types" OFF)
 option(DUMP_VOICES "Dump voices to RIFF WAVE files" OFF)
+option(BUILD_SHARED_LIBS "Build shared library" ON)
 if(WIN32)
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Can we invert this to be “BUILD_STATIC” defaulting to off? This makes it a little more clear that a static library is not the typical usage for FAudio.

(Also, is it possible to detect the system FAudio and only use the static lib when not found? Or is there something we need to provide to help make that possible?)

hcorion:
Unfortunately `BUILD_SHARED_LIBS` 
 is [a builtin CMake variable](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html), so we do need to set it.
If you wanted I could change it to
```cmake
option(BUILD_STATIC "Build statoc library" OFF)
if(BUILD_STATIC)
   set(BUILD_SHARED_LIBS OFF)
else()
   set(BUILD_SHARED_LIBS ON)
endif()
```
I personally find this too verbose and harder to understand, but I don't mind changing it.

flibitijibibo:
Wait, so it's a builtin... but it still needs to be present in the project configs anyway? Does that seem weird to anyone else?

Oh well, no sense adding another layer to something like this, even if it's weird. Thanks for the patch!

PR #180, opened 2020-03-04
Support content type 45 for AudioEninge/SoundBank/WaveBank

alesliehughes:
Game: Drakensang - The River of Time.

### DIFF
--- a/src/FACT_internal.c
+++ b/src/FACT_internal.c
@@ -60,6 +60,7 @@
 #define DOSWAP_64(x) x = SWAP_32(x)
 
 #define FACT_CONTENT_VERSION_3_1 44
+#define FACT_CONTENT_VERSION_3_4 45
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Worth noting that this may also apply to XACT 3.3 - this should be checked with Blazblue Calamity Trigger.

alesliehughes:
Maybe, I currently don't have access to that game, so cant test it directly.

flibitijibibo:
Worst case it just means we have to rename "3_4" to "3_3" for clarity. I've also moved all the version stuff to one spot:

https://github.com/FNA-XNA/FAudio/commit/b5f57cb8787abac9cad612b9c6daacb24e2d98ff

alesliehughes:
Thanks, that look cleaner.

PR #192, opened 2020-05-19
[WIP] Improved gstreamer support

aeikum:
This isn't quite ready to merge, but it's close. With this branch, I get good audio from Skyrim with gst-libav's WMA decoder. Further testing with games you were trying before is appreciated.

First commit is the WMA test I used to verify behavior with xaudio2. I don't think it's important to merge it, but I put it here if you want to run some tests. The important points I discovered are:

-WMA does allow skipping some of the start of the buffer (i.e. PlayBegin is respected).

-WMA does not allow looping arbitrarily. It always loops back to the start and plays to the end, no matter what you set the Loop variables to.

-Having incorrect `pDecodedPacketCumulativeBytes` values does not break decoding. It does limit the total decoded length to the last value, which is how these types of codecs often work (encode whole blocks and truncate the final block to get the right length).

The 2nd commit is your unchanged gstreamer commit. We can merge things down when it's ready to go.

The 3rd commit fixes the build system so it builds correctly in Proton. I was having issues with it finding the gstreamer installation. I'm not very good with CMake, so I probably did this wrong. In particular, the `PkgConfig::` I think requires a later version of CMake than the CMakeList currently requires. Help here is appreciated.

The 4th commit implements the loop behavior I mentioned above.

The final commit rewrites much of the gstreamer algorithm. I won't recap the whole algorithm, but here are some important points:

-WMA is a stateful decoder, so we can't just hop around with random access. We must decode sequentially.

-But, because FAudio requires some amount of random access (mostly that 2 sample "fuck it" quirk), I store both the current and previous decoded chunks. I think this is sufficient.

-We now reset the decoder when looping. I'm not sure whether this is correct, but it makes sense to me.

-We're now using `decodebin` to let gstreamer select the codec for us.

-We limit the full amount of decoded samples to the last `pDecodedPacketCumulativeBytes` value, which I think matches native.

-Instead of using the `pDecodedPacketCumulativeBytes` values we get from the application, we depend on the decoder to give us the actual block sizes, and seek based on that.

There is one ugly hack that I wonder if you have a solution for. I store the actual decoded-samples values mentioned above in the `blockSizes` array, and use that for finding the current buffer requested by FAudio. But, we need to clear and re-size this array when we get a new buffer in. Currently I'm just using the value of the pointer to see if we get a new buffer incoming, but that feels super ugly. Is there some better way to tell this gstreamer code that we have a new buffer and should forget what we know about the old buffer?


flibitijibibo:
Taking a quick look at the buffer pointers being passed, I _think_ it'll be unique... that gets malloc'd for each buffer that gets submitted, even if the user-submitted pointer is reused for each submission:

https://github.com/FNA-XNA/FAudio/blob/master/src/FAudio.c#L2351

So the case where the pointer is the same mostly comes down to a cruel runtime... but, runtimes can be cruel :(

A possible tweak would be to force that pointer to be NULL when the buffer is knowingly swapped in the mixer, so we'd pretty much be reviving this reset function:

https://github.com/FNA-XNA/FAudio/blob/master/src/FAudio_internal.c#L446

Other than that, it all makes sense to me!

aeikum:
Adding that reset function back in looks like the right answer, thanks. I'll do that.

aeikum:
Any thoughts on the CMake changes?

flibitijibibo:
Looks good to me, everything builds fine on flibitBuild!

soredake:
Hope this will be merged before next major proton version!

aeikum:
I've force-pushed an update to my branch. It's nearly ready to merge, but there is one outstanding issue. The game Banished frequently crashes for me with this work. I haven't figured out why yet. I suspect some buffer overrun or use-after-free in the buffers we allocate in the gstreamer code, but that's just a guess. It's on my plate to keep investigating, but if someone has time to dig into this, please feel free.

Banished turns out to be a _very_ good test case for WMA decoding.

flibitijibibo:
Latest push looks good to me. I remember Banished being a crashy mess with the old stuff so I think this is at feature parity anyway? I never did look into why that was. Either way I'll hold off on merging this until 20.08, in case someone figures out what the deal is and it requires something serious to fix.

flibitijibibo:
Going to CC @GloriousEggroll real quick... as far as I can tell this is at least as functional (if not more functional) than the existing WMA backend, so I'm tempted to merge this for 20.08... Banished isn't happy but from what I remember that game was never happy to begin with. GE, do you have a set of games to test this against? If this works for the WMA catalog you have in mind I can go ahead and get this in.

aqxa1:
I forgot to report it, but this PR was causing crashes with Shenmue 1 & 2 (758330) when I tested it, however it's another game that didn't work well with the FFmpeg implementation either (crackling audio).

aeikum:
Yeah I'd recommend against merging it. There's this mysterious crash, and a couple of clumsy leaks and bugs in the version I have in my repo right now. I'll update the branch now, but the crash remains.

Thanks for the note about Shenmue, having another game with the crash might give me a clue that Banished wasn't.

sdroege:
How and where does it crash? Can you share some details?

aeikum:
Hey Sebastian. For me, with Banished and my most recent push, it crashes about half the time immediately after starting a new game. If it doesn't crash there, then it seems to survive without crashing.

soredake:
[steam-758330.zip](https://github.com/FNA-XNA/FAudio/files/4948368/steam-758330.zip)
Shenmue log with proton and this pr, dunno how useful it is.

sdroege:
I mean, can you provide a backtrace of the crash for example? In which piece of code does it happen?

flibitijibibo:
Dug an assertion failure out of the log:

```
(Shenmue.exe:254946): GStreamer-WARNING **: 17:44:32.619: External plugin loader failed. This most likely means that the plugin loader helper binary was not found or could not be run. You might need to set the GST_PLUGIN_SCANNER environment variable if your setup is unusual. This should normally not be required though.
**
ERROR:/home/danet/Templates/FAudio/src/FAudio_gstreamer.c:114:FAudio_GSTREAMER_FillConvertCache: assertion failed: (gst_pad_push( voice->src.gstreamer->srcpad, src ) == GST_FLOW_OK)
```

sdroege:
Ok so an assertion from your code. Can you get a GStreamer debug log with GST_DEBUG=6 and attach it here?

GloriousEggroll:
Warframe originally was a big test case here but they fixed the issue
internally by shipping their own xaudio2_9. Skyrim is usually another test
case, as well as Saints Row 2. Unfortunately I have not had time to go
through these

aqxa1:
I've attached a [GST_DEBUG=6 log](https://github.com/FNA-XNA/FAudio/files/4950653/shenmue_gstreamer.log) for Shenmue.

flibitijibibo:
This is a wild guess, but I think the problem is that `gst_event_new_stream_start` is being passed a constant string value, when it's meant to be unique like all the GstElements are when linking the pad. Maybe the string can be generated with `gst_pad_create_stream_id`?

sdroege:
That's not going to cause a problem here. I'll go through the log this weekend, no time earlier unfortunately. Thanks!

sdroege:
You're not correctly handling (or rather: not at all) GStreamer error messages:

```
0:00:00.054018000 [334m31281[00m     0x7e050660 [33;01mWARN   [00m [00m           decodebin gstdecodebin2.c:4679:gst_decode_bin_expose:<decodebin0>[00m error: no suitable plugins found:
Missing decoder: Windows Media Audio 8 (audio/x-wma, wmaversion=(int)2, bitrate=(int)20000, channels=(int)1, rate=(int)22050, block_align=(int)929, depth=(int)16, codec_data=(buffer)000000001f0000000000000000000000)

0:00:00.054022000 [334m31281[00m     0x7e050660 [37mDEBUG  [00m [00;01;37;41m         GST_MESSAGE gstelement.c:2110:gst_element_message_full_with_details:<decodebin0>[00m start
0:00:00.054031000 [334m31281[00m     0x7e050660 [36mINFO   [00m [00;01;31;47m    GST_ERROR_SYSTEM gstelement.c:2141:gst_element_message_full_with_details:<decodebin0>[00m posting message: Your GStreamer installation is missing a plug-in.
```

You don't have a WMA8 decoder available on your system. It seems like `gstreamer-libav` is not installed (for example), which would provide this via ffmpeg.

aqxa1:
I do have gst-plugins-libav (Gentoo) installed, unless it's missing wma support for some reason. Does ffmpeg itself need to be compiled with WMA support? I can't see any useflags for it.

sdroege:
The ffmpeg plugin is not loaded / found at all according to the logs, so something seems broken on your system in that regard. Does `gst-inspect-1.0 avdec_wmav2` list anything? I can't really help with Gentoo-specific issues.

The main problem here however is that the code in this PR is not handling errors correctly: error messages on the bus need to be handled and errors from pushing a buffer shouldn't cause an assertion but be handled gracefully.

aqxa1:
@sdroege  Here's a [pastebin](https://pastebin.com/qNPbNChB). It does show something at least.

sdroege:
That's a different GStreamer installation than what was used while producing the log above. That one is loading plugins from `/games/steamlib/steamapps/common/Proton 5.0/dist/lib64/gstreamer-1.0/` and there's no `libgstlibav.so` in there or in any of the other paths it is setting:

```
GST_PLUGIN_SYSTEM_PATH set to /games/steamlib/steamapps/common/Proton 5.0/dist/lib64/gstreamer-1.0:/games/steamlib/steamapps/common/Proton 5.0/dist/lib/gstreamer-1.0
```

soredake:
I'm using kubuntu 20.04.

gst-inspect log: [gst-inspect.tar.gz](https://github.com/FNA-XNA/FAudio/files/4981116/gst-inspect.tar.gz)

shenmue log with GST_DEBUG=6: [steam-758330.zip](https://github.com/FNA-XNA/FAudio/files/4981195/steam-758330.zip)

aqxa1:
@sdroege Ah, I see. So presumably I needed to replace libstlibav.so in Proton's installation, and not just libFaudio.

soredake:
After replacing that library sound works, and game does not crash.

aqxa1:
Yeah, that fixed it for me as well. To be clear, the new DLL (64-bit version) goes in `<steam library path>/steamapps/common/Proton 5.0/dist/lib64/gstreamer-1.0/` and not the existing location of libgstlibav.so (which is in lib64 instead).

There's still some pops and clicks, although it seems to be in the music track as well so maybe it's unrelated.

sdroege:
It will probably also work to put it into `~/.local/share/gstreamer-1.0/plugins`, depending on how the Proton GStreamer version is built.

flibitijibibo:
Seems like the original error checking for the plugin got lost - I think that's all we can technically do in FAudio, right? The rest is up to the system to get right, it seems.

Judging by the results it definitely sounds like this is at parity with the old FFmpeg setup though, so once this has error checking I'll pull this in. Again, the original FFmpeg backend isn't technically supported to begin with and this is at least going in with the intention of actually being used in production somewhere.

aeikum:
The missing libav problem is unrelated to the Banished crash I've been trying to pin down.

sdroege:
Nonetheless, please add proper error checking to the code if you want to prevent such things in the future :) That should be handled gracefully somehow by reporting an error, not by running into an assertion. See what I wrote above:

> The main problem here however is that the code in this PR is not handling errors correctly: error messages on the bus need to be handled and errors from pushing a buffer shouldn't cause an assertion but be handled gracefully.

flibitijibibo:
Yeah, the error will definitely manifest in other ways, be it for Proton or vanilla Wine, as the above troubleshooting shows... but yes, the Banished crash is definitely something else. But, Banished is crashy with the current FFmpeg logic, so this is unfortunately nothing new for that game :/

aeikum:
Pushed commits to remove all of the `g_assert`s for gst errors and replace them with proper error handling. Also added a check for bus errors after initial setup, and after each decoding cycle.

flibitijibibo:
Changes lgtm - @aqxa1, would you be able to try the latest revision with the initial setup that broke? This should still work with the setup that fixed the original issue, now we just need to verify that it also breaks correctly (:psyduck:).

aqxa1:
Just tested the updated PR, and no more crashes in the case where libgstlibav.so isn't available.

flibitijibibo:
Excellent! This can be merged if the commits and their messages are good to go.

src/FAudio_gstreamer.c
```diff
@@ -61,26 +61,37 @@ typedef struct FAudioGSTREAMER
...
 	((len) / (sizeof(float) / (voice->src.format->wBitsPerSample / 8)))
 
-static void FAudio_GSTREAMER_RestartDecoder(FAudioGSTREAMER *gstreamer)
+static int FAudio_GSTREAMER_RestartDecoder(FAudioGSTREAMER *gstreamer)
 {
 	GstEvent *event;
 
 	event = gst_event_new_flush_start();
-	g_assert(event);
+	if (!event)
```
sdroege, line 69: This can't fail, and also not the other `gst_event_new_XXX()` below

src/FAudio_gstreamer.c
```diff
@@ -61,26 +61,37 @@ typedef struct FAudioGSTREAMER
...
 	gst_pad_push_event(gstreamer->srcpad, event);
 
 	event = gst_event_new_segment(&gstreamer->segment);
-	g_assert(event);
-	g_assert(gst_pad_push_event(gstreamer->srcpad, event));
+	if (!event)
+	{
+		return 0;
+	}
+	gst_pad_push_event(gstreamer->srcpad, event);
```
sdroege, line 87: That's correct, don't error out when sending events fails.

src/FAudio_gstreamer.c
```diff
@@ -410,19 +458,74 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
 	 */
 	result = (FAudioGSTREAMER*) pSourceVoice->audio->pMalloc(sizeof(FAudioGSTREAMER));
 	FAudio_zero(result, sizeof(FAudioGSTREAMER));
+
 	result->pipeline = gst_pipeline_new(NULL);
-	g_assert(result->pipeline);
+	if (!result->pipeline)
```
sdroege, line 516: This can't fail

src/FAudio_gstreamer.c
```diff
@@ -410,19 +458,74 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
...
+			pSourceVoice->audio,
+			"Unable to create gstreamer pipeline; is %u-bit gstreamer installed?",
+			sizeof(void *) * 8
+		)
+		goto free_without_bin;
+	}
+
 	decoder = gst_element_factory_make("decodebin", NULL);
-	g_assert(decoder);
+	if (!decoder)
```
sdroege, line 527: This however can fail if for some reason decodebin is not available, so correct to check here

src/FAudio_gstreamer.c
```diff
@@ -410,19 +458,74 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
...
+			sizeof(void *) * 8
+		)
+		goto free_without_bin;
+	}
+
 	g_signal_connect(decoder, "pad-added", G_CALLBACK(FAudio_GSTREAMER_NewDecodebinPad), result);
+
 	result->srcpad = gst_pad_new(NULL, GST_PAD_SRC);
-	g_assert(result->srcpad);
+	if (!result->srcpad)
```
sdroege, line 540: This can't fail (but you might want to give a name to the pad, like `"src"`)

src/FAudio_gstreamer.c
```diff
@@ -444,21 +555,64 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
 	);
 
 	decoder_sink = gst_element_get_static_pad(decoder, "sink");
-	g_assert(decoder_sink);
-	g_assert(gst_pad_link(result->srcpad, decoder_sink) == GST_PAD_LINK_OK);
+	if (!decoder_sink)
```
sdroege, line 611: This can't fail

src/FAudio_gstreamer.c
```diff
@@ -444,21 +555,64 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
...
+			"Unable to get link pipeline"
+		)
+		goto free_with_bin;
+	}
 
 	/* send stream-start */
 	event = gst_event_new_stream_start("faudio/gstreamer");
-	g_assert(event);
-	g_assert(gst_pad_push_event(result->srcpad, event));
+	if (!event)
```
sdroege, line 650: This can't fail

src/FAudio_gstreamer.c
```diff
@@ -444,21 +555,64 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
...
+		LOG_ERROR(
+			pSourceVoice->audio,
+			"Unable to create stream_start event; is %u-bit gstreamer installed?",
+			sizeof(void *) * 8
+		)
+		goto free_with_bin;
+	}
+
+	if (!gst_pad_push_event(result->srcpad, event))
+	{
```
sdroege, line 661: If this fails, ignore that. Only later when pushing buffers check the return value

src/FAudio_gstreamer.c
```diff
@@ -493,10 +647,39 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
 		"codec_data",	GST_TYPE_BUFFER, codec_data,
 		NULL
 	);
-	g_assert(caps);
+	if (!caps)
```
sdroege, line 703: This can't fail

src/FAudio_gstreamer.c
```diff
@@ -493,10 +647,39 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
...
+			"Unable to create gstreamer caps; is %u-bit gstreamer installed?",
+			sizeof(void *) * 8
+		)
+		gst_buffer_unref(codec_data);
+		goto free_with_bin;
+	}
 	event = gst_event_new_caps(caps);
-	g_assert(event);
-	g_assert(gst_pad_push_event(result->srcpad, event));
+	if (!event)
```
sdroege, line 714: This can't fail

src/FAudio_gstreamer.c
```diff
@@ -493,10 +647,39 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
...
+		LOG_ERROR(
+			pSourceVoice->audio,
+			"Unable to create gstreamer caps event; is %u-bit gstreamer installed?",
+			sizeof(void *) * 8
+		)
+		gst_caps_unref(caps);
+		gst_buffer_unref(codec_data);
+		goto free_with_bin;
+	}
+	if (!gst_pad_push_event(result->srcpad, event))
```
sdroege, line 725: Same as above, only check when pushing buffers and not here.

src/FAudio_gstreamer.c
```diff
@@ -509,19 +692,71 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
 		"rate",		G_TYPE_INT, pSourceVoice->src.format->nSamplesPerSec,
 		NULL
 	);
-	g_assert(caps);
+	if (!caps)
```
sdroege, line 748: Can't fail

src/FAudio_gstreamer.c
```diff
@@ -102,19 +113,38 @@ static size_t FAudio_GSTREAMER_FillConvertCache(
...
+			voice
+		);
+		gst_buffer_unref(src);
+		return (size_t) -1;
+	}
+
+	if (	gst_pad_push(
+			voice->src.gstreamer->srcpad,
+			src
+		) != GST_FLOW_OK	)
```
sdroege, line 181: `GST_FLOW_EOS` might be a case you don't want to consider an error (end of stream), if this can ever happen in your situation.

src/FAudio_gstreamer.c
```diff
@@ -94,6 +94,48 @@ static int FAudio_GSTREAMER_RestartDecoder(FAudioGSTREAMER *gstreamer)
...
 
+static int FAudio_GSTREAMER_CheckForBusErrors(FAudioVoice *voice)
+{
+	GstBus *bus;
+	GstMessage *msg;
+	GError *err = NULL;
+	gchar *debug_info = NULL;
+
+	bus = gst_pipeline_get_bus(GST_PIPELINE(voice->src.gstreamer->pipeline));
+	if (!bus)
```
sdroege, line 105: This can't fail

src/FAudio_gstreamer.c
```diff
@@ -0,0 +1,850 @@
...
+	gst_pad_push_event(gstreamer->srcpad, event);
+
+	event = gst_event_new_segment(&gstreamer->segment);
+	if (!event)
+	{
+		return 0;
+	}
+	gst_pad_push_event(gstreamer->srcpad, event);
+
+	gst_element_set_state(gstreamer->pipeline, GST_STATE_PLAYING);
```
sdroege, line 89: You want to check the return value here for errors

src/FAudio_gstreamer.c
```diff
@@ -0,0 +1,850 @@
...
+	}
+
+	if (	gst_pad_push(
+			voice->src.gstreamer->srcpad,
+			src
+		) != GST_FLOW_OK	)
+	{
+		LOG_ERROR(
+			voice->audio,
+			"for voice %p, failed to copy whole chunk into buffer",
```
sdroege, line 185: Copy and paste error from above and you probably want to also include the return value of `gst_pad_push()` in the message

src/FAudio_gstreamer.c
```diff
@@ -0,0 +1,850 @@
...
+	GValue gval = G_VALUE_INIT;
+	GstPad *decoder_sink;
+	GstEvent *event;
+
+	LOG_FUNC_ENTER(pSourceVoice->audio)
+
+	/* Init GStreamer statically. The docs tell us not to exit, so I guess
+	 * we're supposed to just leak!
+	 */
+	if (!gstreamerWasInit)
```
sdroege, line 481: Instead of keeping track of it yourself you can also call `gst_is_initialized()`.

src/FAudio_gstreamer.c
```diff
@@ -0,0 +1,850 @@
...
+	uint8_t *extradata;
+	uint8_t fakeextradata[16];
+	GValue gval = G_VALUE_INIT;
+	GstPad *decoder_sink;
+	GstEvent *event;
+
+	LOG_FUNC_ENTER(pSourceVoice->audio)
+
+	/* Init GStreamer statically. The docs tell us not to exit, so I guess
+	 * we're supposed to just leak!
```
sdroege, line 479: If you want you can call `gst_deinit()` at the very end, but you can't call `gst_init()` or any other GStreamer API again afterwards

src/FAudio_gstreamer.c
```diff
@@ -0,0 +1,850 @@
...
+		)
+		goto free_without_bin;
+	}
+
+	/* turn off sync so we can pull data without waiting for it to "play" in realtime */
+	g_value_init(&gval, G_TYPE_BOOLEAN);
+	g_value_set_boolean(&gval, FALSE);
+	g_object_set_property(G_OBJECT(result->dst), "sync", &gval);
+	g_value_set_boolean(&gval, TRUE);
+	g_object_set_property(G_OBJECT(result->dst), "async", &gval);
```
sdroege, line 588:
```suggestion
	g_object_set(G_OBJECT(result->dst), "sync", FALSE, "async", TRUE, NULL);
```

src/FAudio_gstreamer.c
```diff
@@ -0,0 +1,850 @@
...
+		"audio/x-wma",
+		"wmaversion",	G_TYPE_INT, version,
+		"bitrate",	G_TYPE_INT, pSourceVoice->src.format->nAvgBytesPerSec * 8,
+		"channels",	G_TYPE_INT, pSourceVoice->src.format->nChannels,
+		"rate",		G_TYPE_INT, pSourceVoice->src.format->nSamplesPerSec,
+		"block_align",	G_TYPE_INT, pSourceVoice->src.format->nBlockAlign,
+		"depth",	G_TYPE_INT, pSourceVoice->src.format->wBitsPerSample,
+		"codec_data",	GST_TYPE_BUFFER, codec_data,
+		NULL
+	);
```
sdroege, line 702: Instead of unreffing the buffer in all the different return paths below, just unref it directly here after creating the caps

src/FAudio_gstreamer.c
```diff
@@ -0,0 +1,850 @@
...
+	{
+		LOG_ERROR(
+			pSourceVoice->audio,
+			"Unable to create gstreamer caps; is %u-bit gstreamer installed?",
+			sizeof(void *) * 8
+		)
+		gst_buffer_unref(codec_data);
+		goto free_with_bin;
+	}
+	event = gst_event_new_caps(caps);
```
sdroege, line 713: And here you can directly unref the caps after creating the event

sdroege:
Went shortly through the GStreamer-related parts of the code. Mostly minor things

aeikum:
@sdroege Thanks very much for the review. I've pushed an update fixing all of those issues.

There's two outstanding issues I know about. The Banished crash, and, I think we will return a short chunk of silence if the xa2 buffer we are currently decoding doesn't fill the entire buffer that comes in to `FAudio_INTERNAL_DecodeGSTREAMER`. I still recommend against merging it because of these issues, but it's up to Ethan.

flibitijibibo:
Sounds pretty much like #95... bummer that this doesn't fix it, but it sounds like it pretty much nails the existing behavior but without going straight to FFmpeg.

Thanks to everyone for the investigative work and patches! There's still work to do but this should get the ball rolling again on WMA support.

aeikum:
I think it's different from that. For a stupid example, suppose the FAudio engine requests 441 samples from the Decode function, but the xa2 buffer only provides 200 (e.g. end of buffer). We'll then return 200 samples of content and 241 samples of silence. What we should do is begin processing the next xa2 buffer to retrieve the remaining 241 samples from the next buffer. I have some work in this direction, not done yet.

PR #193, opened 2020-07-13
Handle negative zero rt60_ms values.

rbernon:
Dragon Quest Builders 2 audio is broken because is uses a DecayTime of -0.f, and that causes infinite gain and NaN values creeping up the reverb buffer.

This could also be done in some XAPO_PROCESS_BUFFER_PARAMETERS params validation code, or before calling / within DspReverb_SetParameters.

### DIFF
--- a/src/FAudioFX_reverb.c
+++ b/src/FAudioFX_reverb.c
@@ -150,14 +150,7 @@ static inline void DspDelay_Destroy(DspDelay *filter, FAudioFreeFunc pFree)
 
 static inline float DspComb_FeedbackFromRT60(DspDelay *delay, float rt60_ms)
 {
-	float exponent;
-
-	if (rt60_ms == 0)
-	{
-		return 0;
-	}
-
-	exponent = (
+	float exponent = (
 		(-3.0f * delay->delay * 1000.0f) /
 		(delay->sampleRate * rt60_ms)
 	);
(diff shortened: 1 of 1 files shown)


rbernon:
Note that I'm assuming that negative RT60 values also make no sense, but maybe they do. In which case it will need to be updated accordingly.

flibitijibibo:
Sounds like we could use an OnSetParameters implementation for Reverb:

https://github.com/FNA-XNA/FAudio/blob/master/include/FAPOBase.h#L99

But, a possible hack to cover this: What if we change this...

https://github.com/FNA-XNA/FAudio/blob/master/src/FAudioFX_reverb.c#L688

to

```
FAudio_max(params->DecayTime, FAUDIOFX_REVERB_MIN_DECAY_TIME) * 1000.0f`
```

Does this fix it as well?

rbernon:
Yeah it fixes it too, and then it also makes the `rt60_ms == 0` test useless. AFAICS DspComb_FeedbackFromRT60 is only called from there and from `DspCombShelving_Initialize` which uses constant non-zero rt60_ms values from `COMB_DELAYS[i] + STEREO_SPREAD[c]`.

flibitijibibo:
Perfect, do whatever cleanup you want to do on that and I can merge this branch after it's been pushed! We can replace the `max` with `OnSetParameters` later, if it comes up.

PR #201, opened 2020-08-04
Allow decoder functions to return short reads

aeikum:
This one is maybe worth some thought. I think it makes sense to me, but please make sure it fits with how you understand the engine to work.

### DIFF
--- a/src/FAudio_gstreamer.c
+++ b/src/FAudio_gstreamer.c
@@ -285,14 +285,14 @@ static int FAudio_GSTREAMER_DecodeBlock(FAudioVoice *voice, FAudioBuffer *buffer
 	return gstreamer->blockSizes[block] != (size_t) -1;
 }
 
-static void FAudio_INTERNAL_DecodeGSTREAMER(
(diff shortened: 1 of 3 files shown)


flibitijibibo:
It _should_ be a harmless change, but the weird thing about the decoder is that it's supposed to do the math ahead of time, so for a device with a 480-sample period and a buffer that has 420 samples left, it should be going into the decoder requesting 420 samples to begin with. For PCM/ADPCM this works great, and this change effectively is a no-op for those, but for WMA it sounds like it still hits that issue where it doesn't fill in the number of samples that the buffer claims to have (i.e. #95).

Do we have a quick test for this by chance? `testxwma` loads the whole file at once so unfortunately it's a bad test for streams, which this bug affects the most.

aeikum:
Ah, I see what you mean. If there's a problem here, it may be an issue with the calculations in `FAudioSourceVoice_SubmitSourceBuffer`. I'll take a closer look.

aeikum:
Let's just close this for now. I'm not certain there's not a problem here, but at least it hasn't cropped up in the games I've been working on.

PR #202, opened 2020-08-05
Protect caches

aeikum:
I figured out the Banished crash!! The game creates new source voices in one thread while the engine thread is mixing sources. This would cause the decodeCache to be reallocated near the end of FAudio_CreateSourceVoice, while the buffer was still in use (e.g. being written into by the gst decoding code). This patchset moves the source mutex unlocking around each callback invocation, instead of around the entire mixing procedure (see 0a7e525702cd21cfee39b9816a152a1ab25ae538), and then takes the lock when it tries to realloc the buffer.

The third patch was just something I noticed while working on this code; feel free to ignore it if it's not worth the noise.

### DIFF
--- a/src/FAudio_internal.c
+++ b/src/FAudio_internal.c
@@ -636,6 +636,20 @@ static inline void FAudio_INTERNAL_FilterVoice(
 	LOG_FUNC_EXIT(audio)
 }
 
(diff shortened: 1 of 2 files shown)


flibitijibibo:
I just love those callbacks so much~

PR #210, opened 2020-08-25
Use memcpy when we can for mixing

aeikum:
This fixes frequent underruns during cutscenes in Beyond Two Souls on my PC. The game does a ton of 8->8 channel sends with an identity matrix, which we can simply replace with a memcpy. I think we could likely do more improvements in the non-identity case, but I'm not at all familiar with SSE stuff.

### DIFF
--- a/src/FAudio_internal_simd.c
+++ b/src/FAudio_internal_simd.c
@@ -1277,6 +1277,49 @@ void FAudio_INTERNAL_Amplify_NEON(
 #endif /* HAVE_NEON_INTRINSICS */
 
 /* SECTION 4: Mixer Functions */
+static int FAudio_MixerCanFastpath(
+	uint32_t srcChans,
+	uint32_t dstChans,
+	float baseVolume,
+	float *restrict channelVolume,
+	float *restrict coefficients
+) {
+	uint32_t i, j;
+	if (srcChans != dstChans)
+	{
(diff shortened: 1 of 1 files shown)


aeikum:
(Pushed a fix to use FAudio_memcpy instead of memcpy.)

flibitijibibo:
This makes sense to me - I wonder if we can offload this to the API side rather than the mixer? My thinking is that we can do the same check when SetOutputVoices/SetVolume/SetChannelVolumes is called, then all the mixer has to do is check for `isIdentityGain` and go straight to memcpy without calling the function pointer at all. Less maintenance on the SIMD side and possibly performs even better in this case?

aeikum:
I was working on that when I noticed an obvious problem - we need to be _mixing into_ these buffers, not overwriting them. So I'm not sure memcpy is an option at all, here.

aeikum:
I'm going to see if I can do this with SSE2 intrinsics. Let me know if you think this is a bad path to go down.

flibitijibibo:
Oh, duh, you're right... memcpy would only work for the first source voice.

If there's a way we can do it with SIMD that's good enough for me. NEON is pretty easy to port over to as well (at least in my experience), I'll just have to dig up an ARM device to test against if someone doesn't beat me to it.

PR #211, opened 2020-08-31
Provide SSE generic mixing function

aeikum:
The first commit removes a lot of the multiplications from the mixing functions. That speeds things up a little bit, but not very noticeably.

The second commit uses SSE to perform the attenuation, while we have at least four source channels remaining to pull from. It significantly improves performance and fixes the underruns for me in Beyond Two Souls. The game is kind of a special case in that it does a lot of mixing from 8-channel sources, which will hit the fast path very often.


flibitijibibo:
Looks great, thanks for checking this out!

PR #212, opened 2020-09-01
Add workaround for emscripten buffer size requirements

TheSpydog:
Added a check for Emscripten so we can override the `want.samples` value. That said, I don't really understand how this works, so I figured it'd be best if you wrote the comment. 😅


flibitijibibo:
I'll write up a version of this tomorrow, there are a few backends that need this (for now). It'll need a warning message at runtime since the quantum size mismatching can actually cause crashes for specific games.

flibitijibibo:
Nevermind, had a quick minute:

https://github.com/FNA-XNA/FAudio/commit/8d4d9f4d57fc6bf229f3113e0a4a39a5ffd3ca71

Issue #215, opened 2020-09-11
[Regression] 20.09 deadlocks Wolfenstein: The Old Blood


armageddon51:
This new version freeze my wine game (wolfenstein). Observe on 2 pcs (nvidia - amd). Return to version 20.08 solve the issue.

Manjaro XFCE 64 bits.

flibitijibibo:
Pretty sure these are the only two commits that touched locks:

https://github.com/FNA-XNA/FAudio/commit/0b8f62128c7868c47b60d1a2062d6837b186539a
https://github.com/FNA-XNA/FAudio/commit/cea3cddd0fa3bcb0717a89ae936d1e1d61202fb7

CC @aeikum, not sure about the newest game but The New Order is basically just the id5 sound engine, which should be roughly like DOOM 3 BFG's: https://github.com/id-Software/DOOM-3-BFG/tree/master/neo/sound/XAudio2

armageddon51:
I forgot to be more precise, it's Wolfenstein the Old Blood (2015). I don't have any other of the series installed.

flibitijibibo:
@armageddon51 Thanks, should narrow it down a bit. Before Andrew gets a look at this, you will want to run with `WINEDEBUG=+xaudio2` so there are logs to work with.

armageddon51:
I am really not an expert on wine, so where does the "WINEDEBUG=+xaudio2" goes ? Can it be in the command line ?

"env WINEPREFIX="/home/maurice/.wine" /usr/bin/wine C:\\\\windows\\\\command\\\\start.exe /Unix /run/media/maurice/Tforce-data/Jeux/Wolfenstein\ The\ Old\ Blood/WolfOldBlood_x64.exe"

flibitijibibo:
Same idea as WINEPREFIX, so set the variable in the same way and it should produce a pretty big log file. You may want to pipe stdout/stderr to a file.

armageddon51:
Ok not sure where the log file will be but with that variable set "WINEDEBUG=+xaudio2", that stop the freezing. I have been playing for 10 minutes. I tried without been set and it only took a few minute to freeze. So your are on to something.

I have to brush off my Linux piping !

flibitijibibo:
That seems to magnify the race condition then. If you have the output file, post it here.

armageddon51:
So you mean that I have to give you a log file even if there is no freezing with that variable ON ? 

[...]

PR #218, opened 2020-09-14
Resample cache locks

aeikum:
First commit moves allocation of the resample cache into the mixer code, like you suggested (#215).

Second commit mirrors the submix voice mixing, where it holds sendLock through the whole mixing procedure. My motivation here was to protect access to resampleSamples, which otherwise could change mid-mix (I guess it still can via callbacks...).

Please review carefully, there's a lot of locks going on here, and it's possible I made a mistake. One open question is whether we should hold the bufferLock while calling the callbacks. We currently do, so I didn't change that, but I wonder if we should drop the lock?

### DIFF
--- a/src/FAudio.c
+++ b/src/FAudio.c
@@ -2830,20 +2830,16 @@ uint32_t FAudioSourceVoice_SetSourceSampleRate(
 		voice->sends.pSends[0].pOutputVoice->master.inputSampleRate :
 		voice->sends.pSends[0].pOutputVoice->mix.inputSampleRate;
 
-	FAudio_PlatformUnlockMutex(voice->sendLock);
-	LOG_MUTEX_UNLOCK(voice->audio, voice->sendLock)
-
-	/* Resize resample cache */
 	newResampleSamples = (uint32_t) (FAudio_ceil(
 		(double) voice->audio->updateSize *
 		(double) outSampleRate /
 		(double) voice->audio->master->master.inputSampleRate
 	));
-	FAudio_INTERNAL_ResizeResampleCache(
-		voice->audio,
-		newResampleSamples * voice->src.format->nChannels
-	);
 	voice->src.resampleSamples = newResampleSamples;
+
+	FAudio_PlatformUnlockMutex(voice->sendLock);
+	LOG_MUTEX_UNLOCK(voice->audio, voice->sendLock)
+
 	LOG_API_EXIT(voice->audio)
 	return 0;
 }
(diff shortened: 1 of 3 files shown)


aeikum:
A possible improvement would be to re-check the cache size immediately before each usage, instead of just once at the start. That should let us notice if it changed during a callback. But I'll wait to hear what you think.

flibitijibibo:
@armageddon51, @grapeharvest, are you able to test this?

At a glance it's about what I expected, though at this point `sendLock` is almost a lock for the whole voice object, so that may be worth checking out for deadlocks. If this ends up being the case, `bufferLock` may indeed be unnecessary now (we'll probably want to rename `sendLock` to `voiceLock` or something). The optimization sounds like a good idea as well.

grapeharvest:
With just a94d666 I get hard crash after a few seconds and need to kill process.
a94d666 + 94c83e6 a black screen before entering the main menu.

aeikum:
OK I've pushed one more update. This one looks good to me. The third commit is optional, I'm not sure whether we hold that lock intentionally or not during callbacks. It seems wrong to me to hold it.

flibitijibibo:
This all makes sense to me, including the buffer lock - we probably want to hold as few mutexes as humanly possible when jumping into a callback. It does definitely feel like `sendLock` is becoming a whole-voice lock, but if it doesn't affect performance then I can't complain... (admittedly lock performance was only ever noticeable on Switch anyway)

One cast call for testing, mostly to check performance: @grapeharvest @armageddon51

grapeharvest:
Still good for Wolfenstein.

armageddon51:
I really don't mind testing it out but I am not at your Linux knowledge level and all this mumbo-jumbo coding talk. So unless someone have the patience to show me where to get this new file and how to build it, I'll have to wait for the new official release.

flibitijibibo:
Very well - AFAICT this should be what fixes it, according to @grapeharvest's test data. The overall fix is to have as few locks held as possible while callbacks are active.

armageddon51:
Ok, calling back my (very) old programming days and a bit of googling, I download the new sources and build a new "libFAudio.so.0.20.09" file which I replace over the original file directly in the "/usr/lib/" folder (bypassing Pamac). Not sure if I have to replace lib32-FAudio ?

If that's ok, then my Wolfenstein works fine now. No more freezing and performance is as before. So looks good ! Thanks.

flibitijibibo:
Nah, modern Wolfenstein is 64-bit only so that should cover it. Thanks for checking this out (and for the original report)!

PR #221, opened 2020-09-20
Allow Engine Initial to not have global settings

alesliehughes:
This allows xacteninge 2 series to use FAudio.

### DIFF
--- a/src/FACT_internal.c
+++ b/src/FACT_internal.c
@@ -1991,7 +1991,13 @@ uint32_t FACT_INTERNAL_ParseAudioEngine(
 
 	uint8_t *ptr = (uint8_t*) pParams->pGlobalSettingsBuffer;
 	uint8_t *start = ptr;
-	
+
+	/* FIXME: Should be recorded so we can return the correct error */
+	if (!pParams->pGlobalSettingsBuffer)
+	{
+		return 0;
+	}
+
 	uint8_t se = 0; /* Swap Endian */
 	uint32_t magic = read_u32(&ptr, 0);
 	se = magic == 0x58475346;


flibitijibibo:
Minor thing, but: FAudio uses braces for all ifs/loops, including for one-line statements.

Do we have any documentation for XACT 2? Last I checked all of that was private code prior to the 2006 DirectX SDK.

flibitijibibo:
Turns out this is hypothetically possible with XACT3 as well:

https://github.com/hrydgard/minidx9/blob/master/Include/xact3.h#L1519

This is another thing that probably requires a ton of runtime error checking, but I believe we set everything to 0 and the engine will bail out in the form of loops that never iterate. Still, why use a middleware with no content... what were they thinking?

alesliehughes:
Fixed the bracket issue. 

Not that I know of.  I'm just working through games I have with the new xactengine2 patchset in staging.  The are plenty of examples online that never set the global settings variable.

This is the first but the other are going to take more time to implement correctly.

Documentation - Are you using taking about MSDN help? or a PDF that states on how it all works?

flibitijibibo:
I was hoping an xact2.h would be present, but if it exists it's technically a leak, to the best of my knowledge.

Either way, it looks like this is valid in 3.0 as well, so the comment can be removed. We may need a FIXME to point out a lack of error checking for the error code mentioned above, though.

alesliehughes:
Updated again.

xact.h is the header for xactengine 2.0 series.

flibitijibibo:
Here's hoping 2.0->3.0 is little more than the move to XNA...

PR #222, opened 2020-09-30
Correct packing for structure FACTRuntimeParameters

alesliehughes:
No sure how to handle this without breaking things.  Under 64bits the structure size is wrong.

In wine there there is code to copy the structure
memcpy (xact, fact, sizeof(fact)) which wont work when the packing is incorrect.

I can manually copy the structure in wine, which will stop one issue. 

Maybe we just mark it for an update for FAudio version 2 :)

### DIFF
--- a/include/FACT.h
+++ b/include/FACT.h
@@ -101,20 +101,6 @@ typedef void (FACTCALL * FACTNotificationCallback)(
 	const FACTNotification *pNotification
 );
 
-typedef struct FACTRuntimeParameters
-{
-	uint32_t lookAheadTime;
-	void *pGlobalSettingsBuffer;
-	uint32_t globalSettingsBufferSize;
-	uint32_t globalSettingsFlags;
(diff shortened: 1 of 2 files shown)


flibitijibibo:
I can take in the second commit, but yeah, we're stuck with the current layout I'm afraid...

What is the difference between the two struct layouts without the pack(1)? In the comment noting the bug I'd like to have the exact difference so end users know what's wrong if they try to memcpy.

alesliehughes:
The FAudio structure ends up being 8 bytes bigger.  Using memcpy the way you do, it will overflow the ms structure, possible corrupting memory.

flibitijibibo:
Noted it generically for now:

https://github.com/FNA-XNA/FAudio/commit/6b360a33844cc2727c12fd498d01cdae4bad8c46

If it matters we can specify the exact fields that are wrong later.

Issue #224, opened 2020-10-03
FAudio doesn’t generate a pkg-config file


Gcenx:
Compiling from source produces the library along with the needed headers but a pkg-config .pc file is not generated and I don’t see a template within the project.

Is this something you’d be willing to implement?, or is this something you’d rather leave up-to package maintainers.

The only project I’m aware of that checks pkg-config for FAudio currently is wine.

flibitijibibo:
If someone wants to add it to CMakeLists.txt, sure. Nobody's said anything about until now. I strongly suggest looking at everybody's packages and looking for any patches related to this.

ryandesign:
> Nobody's said anything about until now.

Not exactly true. The notion of an FAudio pkg-config file has previously come up in #159.

flibitijibibo:
Looks like their preference was a manual file as well. I noticed that neither Fedora nor Debian has this file, and Arch also has a manual file that looks nothing like Gentoo's. Is this something we _should_ have in upstream? If it's not something that CMake can generate trivially I'm not sure it'd be a good idea.

flibitijibibo:
It looks like we can do an FAudio.pc.in and use configure_file:

https://cmake.org/cmake/help/v3.18/command/configure_file.html

We can can copy SDL2 here that'd be my preference:

https://hg.libsdl.org/SDL/file/1b1fae168557/sdl2.pc.in
https://hg.libsdl.org/SDL/file/1b1fae168557/CMakeLists.txt

Won't make the same thing that others have made, but it'll have to do.

Gcenx:
Arch is Arch, I'd checked the AUR package but not the standard version. The Gentoo package seem to be maintained by a wine contributor so I'd opted to copy what they'd done.

(you beat me to the SDL2 reference)

Having this within upstream is preferable so it can be taken advantage of, as not everyone/system will have headers/libraries within "standard" locations

Also if this does get added and documented, having the headers within /FAudio sub-directory would be ideal.

flibitijibibo:
Just checking on this - I won't be able to write this up for quite a while, but if a packager wants to scribble this up I can review and merge it and CC other maintainers as needed.

flibitijibibo:
Definite 100% ultimate last call for this - FAudio got no commits this month so this would pretty much be the only change we would push, making it a very very safe release to include this in.

PR #228, opened 2020-11-12
cmake: Add a pkgconfig file.

orbea:
Fixes https://github.com/FNA-XNA/FAudio/issues/224.

`JoinPaths.cmake` is modified from jsoncpp.

https://github.com/open-source-parsers/jsoncpp/commit/5be07bdc5e2d5b7715ecbc73749af3e625674dcb

I based the pc file off the copy installed in gentoo.

https://gitweb.gentoo.org/repo/gentoo.git/tree/app-emulation/faudio/files/faudio.pc?id=e0ec6727d673a6ac44c7ee7d82d2ebc609e1f511

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -441,6 +441,20 @@ install(
 	ARCHIVE DESTINATION ${FAudio_INSTALL_LIBDIR}
 )
 
+# Generate a pkgconfig file
+include(cmake/JoinPaths.cmake)
+join_paths(FAUDIO_PKGCONF_LIBDIR "\${prefix}" "${CMAKE_INSTALL_LIBDIR}")
+join_paths(FAUDIO_PKGCONF_INCLUDEDIR "\${prefix}" "${CMAKE_INSTALL_INCLUDEDIR}")
+configure_file(
+	"${PROJECT_SOURCE_DIR}/cmake/${PROJECT_NAME}.pc.in"
+	${PROJECT_BINARY_DIR}/generated/${PROJECT_NAME}.pc
+	@ONLY
+)
+install(
+	FILES ${CMAKE_CURRENT_BINARY_DIR}/generated/${PROJECT_NAME}.pc
+	DESTINATION ${FAudio_INSTALL_LIBDIR}/pkgconfig
+)
+
 # Generate cmake-config file, install CMake files
 include(CMakePackageConfigHelpers)
 configure_package_config_file(
--- a/cmake/FAudio.pc.in
+++ b/cmake/FAudio.pc.in
@@ -1,0 +1,12 @@
+prefix=@CMAKE_INSTALL_PREFIX@
+exec_prefix=${prefix}
+libdir=@FAUDIO_PKGCONF_LIBDIR@
+includedir=@FAUDIO_PKGCONF_INCLUDEDIR@
+
+Name: @PROJECT_NAME@
+URL: https://github.com/FNA-XNA/FAudio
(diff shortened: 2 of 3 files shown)


orbea:
@flibitijibibo This will use the `LIB_VERSION` variable to set the version field in the pkgconfig file which currently is `0.20.11`. However the current release tag is `20.11`. Which is the correct version? I think it would be nice if there was some consistency here.

flibitijibibo:
The soversion should be 0.20.11 - the 0 is effectively the ABI version, so if for some horrendous reason we have to break games referencing libFAudio.so.0 won't be affected after we increment to 1.

In any case, CC @Gcenx - does this patch work for you?

Gcenx:
I’ll test this tonight and let you know

Gcenx:
@flibitijibibo seems to be fine bar the version string within the pc file, didn't check if installing includes to a subdirectories works however.

orbea:
I missed one case where `CMAKE_INSTALL_PREFIX` is equal to `CMAKE_INSTALL_<dir>` and updated the PR for that.

> didn't check if installing includes to a subdirectories works however.

Seems to work here.

With either `-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_INCLUDEDIR=/usr/include/FAudio` or `-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_INCLUDEDIR=include/FAudio` I get the following.
```
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib64
includedir=${prefix}/include/FAudio

Name: FAudio
Description: Accuracy-focused XAudio reimplementation for open platforms
Version: 0.20.11

Libs: -L${libdir} -lFAudio
Cflags: -I${includedir}
```
And with `-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_INCLUDEDIR=/opt/FAudio/include/FAudio`
```
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib64
includedir=/opt/FAudio/include/FAudio

Name: FAudio
Description: Accuracy-focused XAudio reimplementation for open platforms
Version: 0.20.11

Libs: -L${libdir} -lFAudio
Cflags: -I${includedir}
```

Gcenx:
Only thing would be adding `URL:` to the pc file.

The version string starting with 0. could cause problems

orbea:
> The version string starting with 0. could cause problems

What kind of problems are you thinking of? This is rather common practice with other programs that use semver and the pkgconfig file should match the library version rather than the package version. There could be a case for making future release tags match the library version, but its probably not super important?

> Only thing would be adding `URL:` to the pc file.

Good idea, I added this. Here is the default generated pc file with no cmake arguments,.
```
prefix=/usr/local
exec_prefix=${prefix}
libdir=${prefix}/lib64
includedir=${prefix}/include

Name: FAudio
URL: https://fna-xna.github.io/
Description: Accuracy-focused XAudio reimplementation for open platforms
Version: 0.20.11

Libs: -L${libdir} -lFAudio
Cflags: -I${includedir}
```

orbea:
> I’d say the URL should be https://github.com/FNA-XNA/FAudio as to not confuse FNA with FAudio though they are related.

Sounds fair, I changed it.

Gcenx:
@flibitijibibo I think this is now good to go

flibitijibibo:
Excellent, thanks to both of you for checking this out!

PR #229, opened 2020-11-14
Always set OutputFormat.SubFormat to MFAudioFormat_PCM in FAudio_PlatformGetDeviceDetails.

shwnchpl:
Adds a test for the value of the OutputFormat.SubFormat UUID on the results returned from IXAudio27_GetDeviceDetails and updates FAudio_PlatformGetDeviceDetails to always set this field to MFAudioFormat_PCM, mirroring the behavior of native xaudio2.7.  This fixes in-game audio in at least two Ubisoft games (Far Cry Primal and Watch Dogs).

Fixes ValveSoftware/proton#2425

### DIFF
--- a/src/FAudio.c
+++ b/src/FAudio.c
@@ -697,7 +697,8 @@ uint32_t FAudio_CreateMasteringVoice(
 	WriteWaveFormatExtensible(
 		&audio->mixFormat,
 		audio->master->outputChannels,
-		audio->master->master.inputSampleRate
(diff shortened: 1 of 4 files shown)


flibitijibibo:
I always knew the Ubisoft fix would be something ridiculous... excellent catch, thanks for the patch!

shwnchpl:
My pleasure! :grin: And thank you!

PR #232, opened 2020-12-11
Increase verbosity of SDL dependency

rchildre3:
Restore preprocessor macro introduced in:
Commit 7b0973e62c05 ("Warning for SDL <2.0.9")
and removed in:
Commit a0f859c761bb ("Enforce a quantum size measured in millisecon...")
to reduce confusion as to the compiler errors

Include the exact version number in the README and include a few common
distributions that contain at least the requisite version

The previous version contained an #ifdef for a Windows specific
preprocessor directive but it seems to be supported now in MSVC and
other noteworthy compilers:

https://docs.microsoft.com/en-us/cpp/preprocessor/hash-error-directive-c-cpp
https://gcc.gnu.org/onlinedocs/cpp/Diagnostics.html
https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas

### DIFF
--- a/README
+++ b/README
@@ -15,7 +15,7 @@ library via FAudio#, which you can find in the 'csharp/' directory.
 
 Dependencies
 ------------
-FAudio depends solely on SDL 2.0.9+ (Ubuntu 20.04+, Debian 10+, Fedora 30+)
+FAudio depends solely on SDL 2.0.9 or newer.
 FAudio never explicitly uses the C runtime.
 
 Building FAudio
--- a/README
+++ b/README
@@ -15,7 +15,8 @@ library via FAudio#, which you can find in the 'csharp/' directory.
 
 Dependencies
 ------------
-FAudio depends solely on SDL2. FAudio never explicitly uses the C runtime.
+FAudio depends solely on SDL 2.0.9+ (Ubuntu 20.04+, Debian 10+, Fedora 30+)
+FAudio never explicitly uses the C runtime.
 
 Building FAudio
 ---------------
--- a/src/FAudio_platform_sdl2.c
+++ b/src/FAudio_platform_sdl2.c
@@ -28,6 +28,11 @@
 
 #include <SDL.h>
 
+#if SDL_VERSION_ATLEAST(2, 0, 9)
+#else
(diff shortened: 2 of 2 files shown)


README
```diff
@@ -15,7 +15,8 @@ library via FAudio#, which you can find in the 'csharp/' directory.
 
 Dependencies
 ------------
-FAudio depends solely on SDL2. FAudio never explicitly uses the C runtime.
+FAudio depends solely on SDL 2.0.9+ (Ubuntu 20.04+, Debian 10+, Fedora 30+)
```
flibitijibibo, line 18: The SDL version is fine, I'd prefer not to have the distributions named since it's not comprehensive (which is impossible, admittedly).

flibitijibibo:
```suggestion
FAudio depends solely on SDL 2.0.9 or newer.
```

rchildre3: Fair enough, patching now, just got bit today trying to compile this on an Ubuntu 18.04 box.

SveSop:
@rchildre3 
In case you dont want to self-compile SDL2, you can use my ppa for 18.04 (libSDL2 2.0.12)
[https://launchpad.net/~cybermax-dexter/+archive/ubuntu/sdl2-backport](https://launchpad.net/~cybermax-dexter/+archive/ubuntu/sdl2-backport)

I do some updates of the 18.04 libFAudio package, but i fear i am focusing more on keeping lib updated for 20.04 since i do not really have the time to test newer libraries after upgrading to 20.04 myself.

rchildre3:
> @rchildre3
> In case you dont want to self-compile SDL2, you can use my ppa for 18.04 (libSDL2 2.0.12)
> https://launchpad.net/~cybermax-dexter/+archive/ubuntu/sdl2-backport
> 
> I do some updates of the 18.04 libFAudio package, but i fear i am focusing more on keeping lib updated for 20.04 since i do not really have the time to test newer libraries after upgrading to 20.04 myself.

@SveSop Hey! Yes, I did encounter your PPA in my research, I just wanted to compile for myself and figured a bit more verbosity could save others from the frustration I encountered. But thanks for your contributions to the open-source community!

Issue #234, opened 2021-01-08
Race in FACTAudioEngine_Initialize


j-r:
On my (old and slow) system Hyperdimension Neptunia ReBirth1 exits silently (probably catching the crash) pretty soon after initializing (see +xact3 log in linked wine bug).

This problem vanishes when single stepping through FACTAudioEngine_Initialize or randomly sprinkling some FAudio_sleep(1000) into it.

Have no idea where to look now.

flibitijibibo:
Here's where the thread is made: https://github.com/FNA-XNA/FAudio/blob/master/src/FACT.c#L254

The backing FAudio thread should be fine, FACT would surprise me though since there's a massive mutex protecting everything.

j-r:
Thanks! It seems that one sleep between CreateMasteringVoice and PlatformCreateThread is needed. Perhaps the device needs some time to settle down?

flibitijibibo:
Does it still die if you put the sleep after submix creation? FAudio's create functions pretty much all block so it surprises me that it causes any issues to begin with... for reference, the thread is stuck waiting on the mutex that the main thread holds until everything's started:

https://github.com/FNA-XNA/FAudio/blob/master/src/FACT_internal.c#L1640

flibitijibibo:
On the subject of threads I do think I've found something: If the game registers global callbacks and the API thread executes them, we may be running into a situation where the native thread isn't compatible with the Wine thread. This is one reason we had EngineProcedureEXT:

https://github.com/FNA-XNA/FAudio/blob/master/extensions/EngineProcedureEXT.txt

We may need to amend this extension to include an export for the FACT engine thread. If these calls from the log...

```
0024:trace:xact3:IXACT3EngineImpl_RegisterNotification (0x17c6e8)->(0x31f784)
0024:trace:xact3:unwrap_notificationdesc Type 17
0024:trace:xact3:IXACT3EngineImpl_RegisterNotification (0x17c6e8)->(0x31f784)
0024:trace:xact3:unwrap_notificationdesc Type 6
0024:trace:xact3:IXACT3EngineImpl_RegisterNotification (0x17c6e8)->(0x31f784)
0024:trace:xact3:unwrap_notificationdesc Type 3
0024:trace:xact3:IXACT3EngineImpl_RegisterNotification (0x17c6e8)->(0x31f784)
0024:trace:xact3:unwrap_notificationdesc Type 1
```

... have structs these four guys are always NULL...

[...]

PR #238, opened 2021-01-24
Record all notification events.

alesliehughes:
This patch records all the notifications events xact supports.  I've reorder the if/enums to match the order of the NOTIFICATION messages.


src/FACT_internal.h
```diff
@@ -95,11 +95,23 @@ typedef struct FACTDSPPreset
...
 typedef enum FACTNoticationsFlags
 {
-	NOTIFY_CUEDESTROY        = 0x0001,
-	NOTIFY_SOUNDBANKDESTROY  = 0x0002,
-	NOTIFY_WAVEBANKDESTROY   = 0x0004,
-	NOTIFY_WAVEDESTROY       = 0x0008,
-	NOTIFY_WAVESTOP          = 0x0010,
+	NOTIFY_CUEPREPARED           = 0x0001,
+	NOTIFY_CUEPLAY               = 0x0002,
+	NOTIFY_CUESTOP               = 0x0003,
```
flibitijibibo, line 100: Overlapping bitflag, looks like we'll have to bump the enum size.

flibitijibibo (changes requested):
Pretty minor stuff for the most part. This could be shrunk down in a few places but I'll take care of that separately.

flibitijibibo:
Excellent, thanks! On a related note, #234 may be important regarding Wine's interaction with notification events.

PR #241, opened 2021-03-08
More accurate Fast/Slow/SinCos RPC curves

Woflox:
These seem to be the formulas XACT is using, or as close as I can find by comparing the curves visually.

Why is there no sin or cos in SinCos? We may never know. It does make a perfectly circular arc though.


flibitijibibo:
Looks great, thanks for checking this out!

PR #245, opened 2021-03-12
Process filters and effects despite empty send list

### LINKED ISSUE #243: Volume meter on Mixer with empty send list
We are using FAudio for porting an video game that uses XAudio. We discovered a discrepancy between FAudio and XAudio, although I am not sure whether the game is actually following the XAudio specification.

The game uses a dedicated mixer with the VolumeMeter effect attached to measure volume levels. This mixer is completely separate from the main mixer(s) used for playing sound effects. Certain sounds (mainly voice lines) are passed to one of the main mixers as well as this dedicated mixer for volume measurement.

Naturally, this results in the sound to be played twice as loud, since you are feeding it to two mixers in parallel. Setting the volume to 0 for the _VolumeMeter mixer_ does not work as volume levels are calculated after the volume setting of the mixer is applied.

The workaround of the original developers was to initialize this dedicated mixer with an empty send list. Adapted, the code looks like this:

 ```c++
FAudioVoiceSends sendList;
sendList.pSends = NULL;
sendList.SendCount = 0;
FAudio_CreateSubmixVoice(m_pXAudio2, ppDestMixerVoice, m_uiOutputDeviceAudioChannels, m_uiOutputDeviceSampleRate, 0, 0, &sendList, &chain);
```

Apparently this works in XAudio, having the desired effect of being able to measure volume levels without having the mixer actually outputting sound. In FAudio, due to the empty send list, the `process` function of the VolumeMeter effect is never executed.

W4RH4WK:
fix #243

### DIFF
--- a/src/FAudio_internal.c
+++ b/src/FAudio_internal.c
@@ -999,15 +999,6 @@ static void FAudio_INTERNAL_MixSource(FAudioSourceVoice *voice)
 
 sendwork:
 
-	/* Nowhere to send it? Just skip the rest...*/
-	if (voice->sends.SendCount == 0)
-	{
-		FAudio_PlatformUnlockMutex(voice->sendLock);
-		LOG_MUTEX_UNLOCK(voice->audio, voice->sendLock)
-		LOG_FUNC_EXIT(voice->audio)
-		return;
-	}
-
 	/* Filters */
 	if (voice->flags & FAUDIO_VOICE_USEFILTER)
 	{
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Phew, that's a relief - feel free to send the other cpp/ fixes as well!

W4RH4WK:
The other fix has already been merged a long time ago. Just forgot to reset my local master.

flibitijibibo:
Ah, didn't catch that... it's been literal years since I've looked in that folder so I've forgotten what's happened to it 😅

PR #247, opened 2021-03-17
Basic FUNCTIONAL XMA2 support

0x0ade:
Compared to #66, this is no longer "initial" and it's no longer taking stupid shortcuts.

_\*breathes in\*_

- `gst-libav` ships with `avdec_xma2` (citation needed). While we could [wait for every distro under the sun to expose proper sink caps,](https://gitlab.freedesktop.org/gstreamer/gst-libav/-/merge_requests/124) I've taken the creative liberty to add an optional `autoplug-factories` handler with some dynamic mimetype shenanigans for that specific factory.
- I've added `FAudioXMA2WaveFormat` as FFmpeg currently accepts (but doesn't fully respect) XMA2 extra data.
    - XMA2 wave data no longer gets packed into `FAudioWaveFormatExtensible`. Goodbye, broken shortcuts!
    - To handle the multitude of wave formats more easily, I've added `FAudioAnyWaveFormat` as an union of all wave formats.
- The C# bindings for `FAudio_CreateSourceVoice` and `FAudioSourceVoice_SubmitSourceBuffer` have been modified to take multiple wave formats and `FAudioBufferWMA` into account. Given how both are passed by reference and not by value anyway, I'm expecting things to end up fine. Please let me know if I'm overlooking something though, no matter how obvious.
- `FACTWaveBank_Prepare` reconstructs as much of the XMA2 extra data as necessary and documents all the quirks I've encountered for future reference. Maybe a future replacement decoder will verify the accuracy of this.
- The XMA2 seek table uses **sample indices**, unlike the WMA decoded packet cumulative **bytes** table. FACT converts the seek table to the byte index format that everything else expects.
- `nBlockAlign` wasn't enough to feed the decoder with properly sized blocks (in part due to its 16-bit size limitation) - the GStreamer decoder now uses `dwBytesPerBlock` (32-bit) when necessary.
    - ... and who knows if XMA2 _really_ "is quite similar to WMA Pro" in that aspect...
- The XMA2 data end can be misaligned to the block size while still being aligned to the packet size (2kB). FFmpeg standalone doesn't care about this (it always uses the 2kB packet size), but we must handle this properly.
    - Speaking of misaligned ends: `FAudio_GSTREAMER_FillConvertCache` happily read beyond the amount of encoded bytes into unknown territory. This PR changes that to feed undersized end buffers properly, but I don't know if every format expects that. Checking for the format type and handling it accordingly (return early? pad?) is out of the scope of this PR though.
- I've encountered a deadlock or two with the GStreamer decoder along the way, mainly "we need more samples, let's retry infinitely no matter if we're re-decoding the same block or if we're clearly out of data." As with the C# bindings, please let me know of any side effects which I'm not aware of.
- `FACTWaveBank_GetWaveProperties` now asserts when encountering XMA2 because I've yet to encounter a game that calls it and who knows what games expect? If required, I'll follow up with an approximate implementation.
- prolly some other things ive forgotten abt idk

_\*breathes out\*_

Some things are handled outside of this PR, namely a [FACT big endian variation table fix](https://github.com/FNA-XNA/FAudio/commit/e18b700fe2240d37be815de3a9dc3106596df87c) (thanks for merging that ahead of time!) and a FNA PR which I'll follow up with soon.

By the way, **this finally fixes all XMA2 bugs related to #95 which I've encountered.**

### DIFF
--- a/csharp/FAudio.cs
+++ b/csharp/FAudio.cs
@@ -177,6 +177,36 @@ public static class FAudio
 	}
 
 	[StructLayout(LayoutKind.Sequential, Pack = 1)]
+	public struct FAudioXMA2WaveFormat
+	{
+		public FAudioWaveFormatEx wfx;
+		public ushort wNumStreams;
+		public uint dwChannelMask;
+		public uint dwSamplesEncoded;
+		public uint dwBytesPerBlock;
+		public uint dwPlayBegin;
+		public uint dwPlayLength;
+		public uint dwLoopBegin;
+		public uint dwLoopLength;
+		public byte bLoopCount;
+		public byte bEncoderVersion;
+		public ushort wBlockCount;
+	};
+
+	[StructLayout(LayoutKind.Explicit)]
+	public struct FAudioAnyWaveFormat
+	{
+		[FieldOffset(0)]
+		public FAudioWaveFormatEx wfx;
+		[FieldOffset(0)]
+		public FAudioWaveFormatExtensible wfex;
+		[FieldOffset(0)]
+		public FAudioADPCMWaveFormat wfadpcm;
+		[FieldOffset(0)]
+		public FAudioXMA2WaveFormat wfxma2;
+	}
+
+	[StructLayout(LayoutKind.Sequential, Pack = 1)]
 	public unsafe struct FAudioDeviceDetails
 	{
 		public fixed short DeviceID[256]; /* Win32 wchar_t */
@@ -406,6 +436,18 @@ public static class FAudio
 	);
 
 	[DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl)]
+	public static extern uint FAudio_CreateSourceVoice(
+		IntPtr audio, /* FAudio* */
+		out IntPtr ppSourceVoice, /* FAudioSourceVoice** */
+		ref FAudioAnyWaveFormat pSourceFormat,
+		uint Flags,
+		float MaxFrequencyRatio,
+		IntPtr pCallback, /* FAudioVoiceCallback* */
+		IntPtr pSendList, /* FAudioVoiceSends* */
+		IntPtr pEffectChain /* FAudioEffectChain* */
+	);
+
+	[DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl)]
 	public static extern uint FAudio_CreateSubmixVoice(
 		IntPtr audio, /* FAudio* */
 		out IntPtr ppSubmixVoice, /* FAudioSubmixVoice** */
@@ -623,6 +665,13 @@ public static class FAudio
 		IntPtr voice, /* FAudioSourceVoice* */
 		ref FAudioBuffer pBuffer,
 		IntPtr pBufferWMA /* const FAudioBufferWMA* */
+	);
+
+	[DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl)]
+	public static extern uint FAudioSourceVoice_SubmitSourceBuffer(
+		IntPtr voice, /* FAudioSourceVoice* */
+		ref FAudioBuffer pBuffer,
+		ref FAudioBufferWMA pBufferWMA
 	);
 
 	[DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl)]
--- a/csharp/FAudio.cs
+++ b/csharp/FAudio.cs
@@ -193,19 +193,6 @@ public static class FAudio
 		public ushort wBlockCount;
 	};
 
-	[StructLayout(LayoutKind.Explicit)]
-	public struct FAudioAnyWaveFormat
-	{
-		[FieldOffset(0)]
-		public FAudioWaveFormatEx wfx;
-		[FieldOffset(0)]
-		public FAudioWaveFormatExtensible wfex;
-		[FieldOffset(0)]
-		public FAudioADPCMWaveFormat wfadpcm;
-		[FieldOffset(0)]
-		public FAudioXMA2WaveFormat wfxma2;
-	}
-
 	[StructLayout(LayoutKind.Sequential, Pack = 1)]
 	public unsafe struct FAudioDeviceDetails
 	{
@@ -439,7 +426,7 @@ public static class FAudio
 	public static extern uint FAudio_CreateSourceVoice(
 		IntPtr audio, /* FAudio* */
 		out IntPtr ppSourceVoice, /* FAudioSourceVoice** */
-		ref FAudioAnyWaveFormat pSourceFormat,
+		IntPtr pSourceFormat, /* FAudioWaveFormatEx* */
 		uint Flags,
 		float MaxFrequencyRatio,
 		IntPtr pCallback, /* FAudioVoiceCallback* */
--- a/csharp/FAudio.cs
+++ b/csharp/FAudio.cs
@@ -176,8 +176,7 @@ public static class FAudio
 		 */
 	}
 
-	[StructLayout(LayoutKind.Sequential, Pack = 1)]
-	public struct FAudioXMA2WaveFormat
+	public struct FAudioXMA2WaveFormatEx
 	{
 		public FAudioWaveFormatEx wfx;
 		public ushort wNumStreams;
--- a/include/FAudio.h
+++ b/include/FAudio.h
@@ -186,9 +186,6 @@ typedef struct FAudioADPCMWaveFormat
 	#endif
 } FAudioADPCMWaveFormat;
 
-/* See DirectXTK, MIT-licensed:
- * https://github.com/microsoft/DirectXTK/blob/6b395cc7f4203ddae40314433f79c6da92fa7f08/XWBTool/xwbtool.cpp#L67
- */
 typedef struct FAudioXMA2WaveFormat
 {
 	FAudioWaveFormatEx wfx;
@@ -204,15 +201,6 @@ typedef struct FAudioXMA2WaveFormat
 	uint8_t  bEncoderVersion;
 	uint16_t wBlockCount;
 } FAudioXMA2WaveFormat;
-
-typedef union FAudioAnyWaveFormat
-{
-	FAudioWaveFormatEx;
-	FAudioWaveFormatEx wfx;
-	FAudioWaveFormatExtensible wfex;
-	FAudioADPCMWaveFormat wfadpcm;
-	FAudioXMA2WaveFormat wfxma2;
-} FAudioAnyWaveFormat;
 
 typedef struct FAudioDeviceDetails
 {
@@ -617,7 +605,7 @@ FAUDIOAPI void FAudio_UnregisterForCallbacks(
 FAUDIOAPI uint32_t FAudio_CreateSourceVoice(
 	FAudio *audio,
 	FAudioSourceVoice **ppSourceVoice,
-	const FAudioAnyWaveFormat *pSourceFormat,
+	const FAudioWaveFormatEx *pSourceFormat,
 	uint32_t Flags,
 	float MaxFrequencyRatio,
 	FAudioVoiceCallback *pCallback,
--- a/include/FAudio.h
+++ b/include/FAudio.h
@@ -186,6 +186,34 @@ typedef struct FAudioADPCMWaveFormat
 	#endif
 } FAudioADPCMWaveFormat;
 
+/* See DirectXTK, MIT-licensed:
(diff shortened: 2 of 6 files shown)


0x0ade:
With all that being said: this still isn't perfect.
- [Death vs. The IRS **PC**](https://files.0x0ade.ga/xma2/death_vs_the_irs_pc.wav)
- [Death vs. The IRS **X360**](https://files.0x0ade.ga/xma2/death_vs_the_irs_x360.wav)

To make sure that this isn't the XMA2 assets sounding absolutely horrible, I went ahead and extracted the two songs in full and converted them using FFmpeg.
- [Menu](https://files.0x0ade.ga/xma2/death_vs_the_irs_menu.wav)
- [Level](https://files.0x0ade.ga/xma2/death_vs_the_irs_level.wav)

Any idea as to what could be causing that weird filter-like sound? A stupid thought I've got right now is "sample rate" but that can't be it... yet given all the past "that can't be it"s, it wouldn't fully surprise me given how weird looking the sample rates can be and how FFmpeg rounds them.

src/FAudio.c
```diff
@@ -355,6 +351,31 @@ uint32_t FAudio_CreateSourceVoice(
...
 		(*ppSourceVoice)->src.format = &fmtex->wfx;
 	}
+	else if (pSourceFormat->wFormatTag == FAUDIO_FORMAT_XMAUDIO2)
+	{
+		FAudioXMA2WaveFormat *fmtex = (FAudioXMA2WaveFormat*) audio->pMalloc(
+			sizeof(FAudioXMA2WaveFormat)
+		);
+
+		/* Copy what we can, ideally the sizes match! */
+		size_t cbSize = sizeof(FAudioWaveFormatEx) + pSourceFormat->cbSize;
```
flibitijibibo, line 361: Ditto for VS2010.

src/FAudio.c
```diff
@@ -2451,6 +2483,11 @@ uint32_t FAudioSourceVoice_SubmitSourceBuffer(
 				fmtex->wSamplesPerBlock
 			) - playBegin;
 		}
+		else if (voice->src.format->wFormatTag == FAUDIO_FORMAT_XMAUDIO2)
+		{
+			FAudioXMA2WaveFormat *fmtex = (FAudioXMA2WaveFormat*) voice->src.format;
```
flibitijibibo, line 2488: More VS2010...

src/FAudio_gstreamer.c
```diff
@@ -449,11 +469,60 @@ static void FAudio_GSTREAMER_NewDecodebinPad(GstElement *decodebin,
...
+/* XMA2 doesn't have a recognized mimetype and so far only libav's avdec_xma2 exists for XMA2.
+ * Things can change and as of writing this patch, things are actively changing.
+ * As for this being a possible leak: using gstreamer is a static endeavour elsewhere already~
+ * -ade
+ */
+static const char* FAudio_GSTREAMER_XMA2_Mimetype = NULL;
+static GstElementFactory* FAudio_GSTREAMER_XMA2_DecoderFactory = NULL;
+static GValueArray *FAudio_GSTREAMER_XMA2_DecodebinAutoplugFactories(GstElement *decodebin,
+		GstPad *pad, GstCaps *caps, gpointer user)
+{
```
flibitijibibo, line 481:
```suggestion
static GValueArray *FAudio_GSTREAMER_XMA2_DecodebinAutoplugFactories(
	GstElement *decodebin,
	GstPad *pad,
	GstCaps *caps,
	gpointer user
) {
```

src/FAudio_gstreamer.c
```diff
@@ -473,15 +542,42 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
 		gst_init(NULL, NULL);
 	}
 
+	if (type == FAUDIO_FORMAT_XMAUDIO2 && !FAudio_GSTREAMER_XMA2_Mimetype)
+	{
+		/* Old versions ship with unknown/unknown as the sink caps mimetype.
+		 * A patch has been submitted (and merged!) to expose avdec_xma2 as audio/x-xma with xmaversion 2:
+		 * https://gitlab.freedesktop.org/gstreamer/gst-libav/-/merge_requests/124
+		 * For now, try to find avdec_xma2 if it's found, match the mimetype on the fly.
+		 * This should also take future alternative XMA2 decoders into account if avdec_xma2 is missing.
```
flibitijibibo, line 551: May be good to add a note for what we can discard later on when we can assume that the patch is in active use.

src/FAudio_gstreamer.c
```diff
@@ -607,11 +707,12 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
...
 	/* Prepare the input format */
+	result->blockAlign = (uint32_t) pSourceVoice->src.format->nBlockAlign;
 	if (type == FAUDIO_FORMAT_WMAUDIO3)
 	{
-		const FAudioWaveFormatExtensible *wfx =
-			(FAudioWaveFormatExtensible*) pSourceVoice->src.format;
-		extradata = (uint8_t*) &wfx->Samples;
+		const FAudioWaveFormatExtensible* wfx =
+			(FAudioWaveFormatExtensible*)pSourceVoice->src.format;
+		extradata = (uint8_t*)&wfx->Samples;
```
flibitijibibo, line 715: Looks like style got mangled here...

src/FAudio_gstreamer.c
```diff
@@ -622,22 +723,31 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
 		extradata = fakeextradata;
 		codec_data_size = sizeof(fakeextradata);
 	}
+	else if (type == FAUDIO_FORMAT_XMAUDIO2)
+	{
+		const FAudioXMA2WaveFormat* wfx =
+			(FAudioXMA2WaveFormat*)pSourceVoice->src.format;
+		extradata = (uint8_t*)&wfx->wNumStreams;
```
flibitijibibo, line 730:
```suggestion
		const FAudioXMA2WaveFormat *wfx =
			(FAudioXMA2WaveFormat*) pSourceVoice->src.format;
		extradata = (uint8_t*) &wfx->wNumStreams;
```

src/FAudio_gstreamer.c
```diff
@@ -622,22 +723,31 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
...
-		"depth",	G_TYPE_INT, pSourceVoice->src.format->wBitsPerSample,
-		"codec_data",	GST_TYPE_BUFFER, codec_data,
+		mimetype,
+		versiontype, G_TYPE_INT, version,
+		"bitrate", G_TYPE_INT, pSourceVoice->src.format->nAvgBytesPerSec * 8,
+		"channels", G_TYPE_INT, pSourceVoice->src.format->nChannels,
+		"rate", G_TYPE_INT, pSourceVoice->src.format->nSamplesPerSec,
+		"block_align", G_TYPE_INT, result->blockAlign,
+		"depth", G_TYPE_INT, pSourceVoice->src.format->wBitsPerSample,
+		"codec_data", GST_TYPE_BUFFER, codec_data,
```
flibitijibibo, line 750: Spacing got mangled here.

flibitijibibo (changes requested):
Functionally I'll trust that what you have works. A lot of the notes just concern compatibility with existing data/applications and some style stuff here and there.

For WMA however I'll have to defer to someone else - once everything else is finished I'll CC them in so they can check it out.

src/FAudio_gstreamer.c
0x0ade: The discardable stuff can now be skipped by defining `FAUDIO_GST_LIBAV_EXPOSES_XMA2_CAPS_IN_CURRENT_YEAR`, hopefully that's good enough until the time comes to make that the default 😅

csharp/FAudio.cs
```diff
@@ -176,6 +176,23 @@ public struct FAudioADPCMWaveFormat
 		 */
 	}
 
+	[StructLayout(LayoutKind.Sequential, Pack = 1)]
```
flibitijibibo, line 179: We got confirmation that this is _not_ packed. Kind of weird, but oh well...

include/FAudio.h
```diff
@@ -186,6 +186,22 @@ typedef struct FAudioADPCMWaveFormat
 	#endif
 } FAudioADPCMWaveFormat;
 
+typedef struct FAudioXMA2WaveFormat
```
flibitijibibo, line 189: Have to move this out of the pragma pack now :(

src/FAudio_gstreamer.c
```diff
@@ -607,11 +721,12 @@ uint32_t FAudio_GSTREAMER_init(FAudioSourceVoice *pSourceVoice, uint32_t type)
...
 
 	/* Prepare the input format */
+	result->blockAlign = (uint32_t) pSourceVoice->src.format->nBlockAlign;
 	if (type == FAUDIO_FORMAT_WMAUDIO3)
 	{
 		const FAudioWaveFormatExtensible *wfx =
-			(FAudioWaveFormatExtensible*) pSourceVoice->src.format;
-		extradata = (uint8_t*) &wfx->Samples;
+			(FAudioWaveFormatExtensible*)pSourceVoice->src.format;
+		extradata = (uint8_t*)&wfx->Samples;
```
flibitijibibo, line 729:
```suggestion
			(FAudioWaveFormatExtensible*) pSourceVoice->src.format;
		extradata = (uint8_t*) &wfx->Samples;
```

flibitijibibo (changes requested):
Pass 2, pretty close to ready at this point.

flibitijibibo (approved):
Latest looks good! Going to CC @aeikum to make sure we didn't screw up the WMA support.

aeikum:
Thanks for checking. Unfortunately 810ffd4 crashes both Banished and Castle Crashers for me, so looks like something is going wrong :( I've attached a log here, but it doesn't look super informative to me. The crash is on line 1669.

[steam-204360.txt](https://github.com/FNA-XNA/FAudio/files/6165528/steam-204360.txt)

src/FAudio_gstreamer.c
```diff
@@ -241,7 +262,7 @@ static int FAudio_GSTREAMER_DecodeBlock(FAudioVoice *voice, FAudioBuffer *buffer
 	uint8_t *tmpBuf;
 	size_t tmpLen;
 
-	if (gstreamer->curBlock != ~0u && block != gstreamer->curBlock + 1)
+	if (gstreamer->curBlock != ~0u && block != gstreamer->curBlock && block != gstreamer->curBlock + 1)
```
flibitijibibo, line 265: We're reasonably certain this is the breaking change... undoing this may fix WMA support?

aeikum: Still crashes after reverting that one line.

0x0ade:
In my attempt to decouple XMA2 decoding from `FAudioBufferWMA`, I mistakenly moved an access to it from `FAudio_INTERNAL_DecodeGSTREAMER` to `FAudio_GSTREAMER_init`, which is callable before that buffer was available. I've now changed `DecodeGSTREAMER`'s behavior to either use a value supplied by `init` or to fall back to `bufferWMA`, fixing a crash with a game I was able to test this with locally. Hopefully this also fixes the crashes you were experiencing.

As for the `block != gstreamer->curBlock` change - it was added during development with another change that prevented certain deadlocks (which I'll attach a comment to shortly). Should I revert those changes?

src/FAudio_gstreamer.c
```diff
@@ -404,7 +436,8 @@ static void FAudio_INTERNAL_DecodeGSTREAMER(
 	if (samples > 0)
 	{
 		if (	!error &&
-			curBlock < voice->src.bufferList->bufferWMA.PacketCount - 1	)
+			curBlock < blockCount - 1 &&
+			availableBytes != 0	)
```
0x0ade, line 440: `availableBytes != 0` is the other change I've added to prevent decoding from looping infinitely, specifically when the decoder thinks there are more blocks but ran out of data to decode. In my case happened due to block size problems which have now been fixed. Is this still worth keeping?

flibitijibibo: If your stuff still works without the change I would revert it - for this stuff in particular, the less changes we have the better since this file can get pretty sensitive (as you've seen already).

0x0ade:
@aeikum can you please re-review the PR with the latest two commits? 😅

aeikum:
@0x0ade Confirmed it's no longer crashing, and audio is functional, with 31ae7861c29ae92976409ff289fab4631155ce91.

PR #251, opened 2021-03-21
More X360 content endianness fixes

0x0ade:
Verified using CastleMiner Z for X360 and for PC.


src/FACT_internal.c
```diff
@@ -2320,8 +2321,9 @@ void FACT_INTERNAL_ParseTrackEvents(
 			track->events[i].wave.angle = read_u16(ptr, se);
 
 			/* Track Variation */
-			track->events[i].wave.complex.trackCount = read_u16(ptr, se);
-			track->events[i].wave.complex.variation = read_u16(ptr, se);
+			evtInfo = read_u32(ptr, se);
```
flibitijibibo, line 2324: Probably want to do this for the other sections marked Track Variation, otherwise good

0x0ade: I wanted to avoid applying too many changes without games to test them with but sure enough :)

PR #252, opened 2021-05-21
Add support for client-provided platform callbacks.

rbernon:
Even with the engine procedure callback, Wine has to swap it to another
thread in order for the application-provided callbacks to be executed in
the context of a Wine thread, and not the SDL audio created thread.

This adds unnecessary complexity, where we can simply provide some
platform callbacks to create threads and synchronization primitives
directly instead.

This will also allow Wine get rid of the FAudio SDL2 dependency, which
is desirable, considering we'll ultimately need FAudio built as PE.

@aeikum told me that this was discussed before and the current
approach was preferred at the time, but maybe now that the API is more
stable it can be reconsidered?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ project(FAudio C)
 
 # Options
 option(GSTREAMER "Enable GStreamer support (WMA, XMA)" OFF)
+option(PLATFORM_CALLBACKS "Enable Platform callbacks instead of SDL2" OFF)
 option(BUILD_UTILS "Build utils/ folder" OFF)
 option(BUILD_TESTS "Build tests/ folder for unit tests to be executed on the host against FAudio" OFF)
 if(WIN32)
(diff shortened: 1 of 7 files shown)


flibitijibibo:
Believe it or not I was actually planning to break the internal API again really soon, mainly for FAudioCreateMasteringVoice8 compatibility.

The internal stuff still changes a fair bit, which is saying something considering how rarely FAudio changes in general these days. Usually when there's a change, it's somewhere in the platform layer.

That said, if the plan is to make a PE for Wine, maybe we can revive the WASAPI backend instead? That way it's doing basically what real XAudio2 does anyway, and will let us keep breaking stuff internally as needed. I had the code laid out like new platforms might get added, but at this point I think that's no longer necessary; we've officially hit every single game platform with SDL as of this month (PlayStation was the last one) so I think we can replace "UNKNOWN" with "WASAPI" now.

rbernon:
Sure, it would be even better if there's an upstream Windows-only port. How would the GStreamer part work in that case? Would it be completely disabled or could it use some Windows API to do the same thing?

With this platform callback approach I thought we it could use some additional callbacks to plug it to Wine GStreamer wrapper, but if FAudio can have a pure Windows backend it would actually be perfect (and much easier) for Wine.

flibitijibibo:
How the GStreamer stuff works is totally up to the Wine team - we don't ship it by default for FNA and only use it for research at the absolute most, so the only official consumer is basically Wine (be it from distributions or Valve), so you have total control over how that all works.

rbernon:
The Win32 port is actually much cleaner, and much better as it allows us to have WMA/WMV support through Media Foundation without too much hassle, so I'm closing this.

PR #253, opened 2021-06-10
Add a native Win32 platform implementation. 

rbernon:
As discussed in https://github.com/FNA-XNA/FAudio/pull/252 this is going to be useful for Wine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,7 @@ add_library(FAudio
 if(PLATFORM_WIN32)
 	target_link_libraries(FAudio PRIVATE -ldxguid -luuid -lwinmm -lole32 -ladvapi32 -luser32 -lmfplat -lmfreadwrite -lmfuuid -lpropsys)
 	target_compile_definitions(FAudio PUBLIC FAUDIO_WIN32_PLATFORM)
+	target_compile_definitions(FAudio PRIVATE HAVE_WMADEC=1)
 	set(PLATFORM_CFLAGS "-DFAUDIO_WIN32_PLATFORM")
 	set(GSTREAMER OFF)
 	set(XNASONG OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,7 +153,7 @@ if(GSTREAMER)
 	set(FAUDIO_GSTREAMER_DEPS " gstreamer-app-1.0 gstreamer-audio-1.0")
 
 	# Add the extra definition...
-	target_compile_definitions(FAudio PRIVATE HAVE_GSTREAMER=1)
+	target_compile_definitions(FAudio PRIVATE HAVE_WMADEC=1 HAVE_GSTREAMER=1)
 
 	# Find GStreamer
 	find_package(PkgConfig)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,7 +106,7 @@ add_library(FAudio
 )
 
 if(PLATFORM_WIN32)
-	target_link_libraries(FAudio PRIVATE -ldxguid -luuid -lwinmm -lole32 -ladvapi32 -luser32)
+	target_link_libraries(FAudio PRIVATE -ldxguid -luuid -lwinmm -lole32 -ladvapi32 -luser32 -lmfplat -lmfreadwrite -lmfuuid -lpropsys)
 	target_compile_definitions(FAudio PUBLIC FAUDIO_WIN32_PLATFORM)
 	set(PLATFORM_CFLAGS "-DFAUDIO_WIN32_PLATFORM")
 	set(GSTREAMER OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ option(BUILD_UTILS "Build utils/ folder" OFF)
 option(BUILD_TESTS "Build tests/ folder for unit tests to be executed on the host against FAudio" OFF)
 if(WIN32)
 option(BUILD_CPP "Build cpp/ folder (COM wrapper for XAudio2)" OFF)
+option(PLATFORM_WIN32 "Enable native Win32 platform instead of SDL2" OFF)
 endif()
 option(XNASONG "Build with XNA_Song.c" ON)
 option(LOG_ASSERTIONS "Bind FAudio_assert to log, instead of platform's assert" OFF)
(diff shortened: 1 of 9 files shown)


src/XNA_Song.c
```diff
@@ -25,11 +25,13 @@
  */
 
 #ifndef DISABLE_XNASONG
+#ifndef FAUDIO_WIN32_PLATFORM
```
flibitijibibo, line 28: Probably easier to disable XNA_Song for Win32 configurations

src/FAudio_platform_sdl2.c
```diff
@@ -26,6 +26,7 @@
 
 #include "FAudio_internal.h"
 
+#ifndef FAUDIO_WIN32_PLATFORM
```
flibitijibibo, line 29: Can go above FAudio_internal.h, to avoid preprocessing extra stuff (may need an `extern int this_tu_is_empty` like we do in other files)

src/FAudio_gstreamer.c
```diff
@@ -26,6 +26,7 @@
  */
 
 #ifdef HAVE_GSTREAMER
+#ifndef FAUDIO_WIN32_PLATFORM
```
flibitijibibo, line 29: Probably easier to disable GStreamer for Win32 configurations

CMakeLists.txt
```diff
@@ -98,11 +99,21 @@ add_library(FAudio
...
+	src/FAudio_platform_win32.c
 	# Optional source files
 	src/XNA_Song.c
 	src/FAudio_gstreamer.c
 )
 
+if(PLATFORM_WIN32)
+	target_link_libraries(FAudio PRIVATE -ldxguid -luuid -lwinmm -lole32 -ladvapi32 -luser32 -lmfplat -lmfreadwrite -lmfuuid -lpropsys)
+	target_compile_definitions(FAudio PUBLIC FAUDIO_WIN32_PLATFORM)
+	target_compile_definitions(FAudio PRIVATE HAVE_GSTREAMER=1)
```
flibitijibibo, line 111: Slightly conflicts with what happens in gstreamer.c

src/FAudio_internal.h
```diff
@@ -28,19 +28,21 @@
...
-#ifdef FAUDIO_UNKNOWN_PLATFORM
+#ifdef FAUDIO_WIN32_PLATFORM
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <math.h>
 #include <assert.h>
 #include <inttypes.h>
 
+#include <windows.h>
```
flibitijibibo, line 39: Seems like a pretty heavy include for the library, is this needed everywhere or just platform_win32.c?

src/FAudio_platform_win32.c
```diff
@@ -0,0 +1,1414 @@
...
+ *
+ * Ethan "flibitijibibo" Lee <flibitijibibo@flibitijibibo.com>
+ *
+ */
+
+#include "FAudio_internal.h"
+
+#include <stddef.h>
+
+#ifdef FAUDIO_WIN32_PLATFORM
```
flibitijibibo, line 31: Can go above FAudio_internal.h, similar to platform_sdl2.c

flibitijibibo (changes requested):
A lot of this is mostly just to make it so both the default and Win32 builds do as little as possible and avoids possible oddities with the Win32 configuration (i.e. somebody tries to enable GStreamer for Win32 builds). As far as win32.c goes I'll just take your word for it that this works.

Aside from the comments this is okay!

src/FAudio_internal.h
rbernon: Yeah I didn't spend much time on it, it's for `lstrcpynA` which lives in `winbase.h` but including just that header made it fail to build. I'll try harder to figure the correct minimal includes for that.

CMakeLists.txt
rbernon: Probably some commit split leftovers, I'll make sure they make more sense individually.

BTW I kept the name GStreamer, but as it's later implemented on top of MF, it's a bit awkward, it may be worth renaming the related code to something like WMADEC, and enable it when either GStreamer is found, or on Win32 platform.

flibitijibibo: The WMADEC option makes sense to me - for Win32 it can be built-in while for other targets it'll require GStreamer. We _might_ need to keep the `GSTREAMER` CMake option for compatibility, but from what I've seen packagers have been really good about changes to the build system since our changesets are so small anyway, so it may not be a huge deal.

rbernon: I kept the `GSTREAMER` CMake flag as it makes sense if someone wants to check for GStreamer availability, and eventual WMADEC support. The win32 platform has it enabled by default.

rbernon:
So I pretty much rewritten the WMADEC part, which was pretty broken, and did some more testing without seeing any trouble for several games. I tried Castle Crashers, Banished, Fallout 4, and they worked nicely with the new MF based WMADEC.

I also tried testxwma.exe with FFmpeg `ergon.xwma` sample, and I used it on windows to check that the use of the MF API was correct. It kind of works there, at least from the pipeline point of view, samples get in and out, but the WMA decoder doesn't work properly. I think the fake private codec data we use for WMAUDIO2 stream, which works fine with FFmpeg is not correct for native MF, and I couldn't figure how to get it right. The `ergon.xwma` sample doesn't have any additional bytes in its format, so I don't really know where it's supposed to come from.

That said, I'd like to test a bit more the other format versions, but I don't know games which use them, or where to get audio samples in such format. Would you have any hint maybe?

For the XNA Song part, I tried a few XNA games as well as a simple executable which calls the XNA API api directly to play random files. On Windows It worked for audio files in various formats but there again, I was not able to play `ergon.xwma` because MF failed to detect its format. I can try harder to make it work on Windows if it's important but I'm afraid I'll hit the same wall when no private codec data is provided.

rbernon:
Please, note that the `WMA` support in `Wine` also needs patches that are not yet upstream (currently available in `Proton`) as well as a few additional ones to expose a `WMA` decoder (for the `WMADEC` part, `XNA Song` uses a simple file based pipeline, which is currently implemented with decodebin and should work already in `Proton`).

flibitijibibo (approved):
ergon.xwma is more in line with what would get sent via SubmitSourceBuffer rather than XNA_Song - if the .wma files in XNA games work you're all set, it's all processed through the same content pipeline. I don't think it ever worked so I'm not too worried about that file... as long as real data works it should be okay.

Latest revision looks great, so I'll merge this as soon as everyone else is ready for me to hit the buttom!

CMakeLists.txt
```diff
@@ -141,7 +154,7 @@ if(GSTREAMER)
 	set(FAUDIO_GSTREAMER_DEPS " gstreamer-app-1.0 gstreamer-audio-1.0")
 
 	# Add the extra definition...
-	target_compile_definitions(FAudio PRIVATE HAVE_GSTREAMER=1)
+	target_compile_definitions(FAudio PRIVATE HAVE_WMADEC=1)
```
flibitijibibo, line 157: Oh wait just realized, this probably has to define _both_, right?

flibitijibibo: Quick ping for
```suggestion
	target_compile_definitions(FAudio PRIVATE HAVE_GSTREAMER=1 HAVE_WMADEC=1)
```
Everything else looks fine so I can merge this right away, and the next release can label this as "experimental" until Wine's ready to go full PE for the XAudio2 implementation.

rbernon: Sure, sorry I've been side tracked a bit. I squashed the fix and updated the PR.

flibitijibibo:
We're all set, thanks for checking this out!

rbernon:
Awesome, thanks!

PR #255, opened 2021-06-22
Don't assert that F3DAudioCalculate channel count match instance count.

rbernon:
This assertion is triggered by Guilty Gear Strive, after it has
called X3DAudioInitialize with a channel mask set to 3, and then calls
X3DAudioCalculate with DstChannelCount set to 6.

A simple program which does the same runs fine on Windows. I'm not completely sure to see why this assertion is there, as the instance channel count doesn't seem to be used anywhere else.

### DIFF
--- a/src/F3DAudio.c
+++ b/src/F3DAudio.c
@@ -311,7 +311,7 @@ F3DAUDIOAPI uint8_t F3DAudioCheckCalculateParams(
 	uint32_t Flags,
 	F3DAUDIO_DSP_SETTINGS *pDSPSettings
 ) {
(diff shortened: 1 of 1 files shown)


flibitijibibo:
This may be a genuine bug in the game - the DirectX SDK has a debug X3DAudio1_7D.dll with assertions enabled, if you rename it and drop it into the game directory and start the game with Visual Studio it may trigger a breakpoint! Skyrim is like this too.

rbernon:
Okay, I'll try that. Maybe it gets the channel mask / count from somewhere where Wine reports inconsistent numbers.

PR #256, opened 2021-06-23
Fix incorrect value for MF_MT_AUDIO_AVG_BYTES_PER_SECOND.

rbernon:
This fixes WMAV2 which was outputting garbage on Windows. I can now play `ergon.xwma` using `testxwma.exe` as well as other WMAV2 streams.


flibitijibibo:
Shoutout to WAVEFORMATEX 🙄

PR #257, opened 2021-07-02
Support WaveBank Entry Names

alesliehughes:


include/FACT.h
```diff
@@ -193,6 +193,7 @@ typedef struct FACTWaveBankEntry
 	FACTWaveBankMiniWaveFormat Format;
 	FACTWaveBankRegion PlayRegion;
 	FACTWaveBankSampleRegion LoopRegion;
+	char szWaveBankName[64];
```
flibitijibibo, line 196: This isn't part of xact3wb as far as I know, so it'll have to be stored a different way. We can probably malloc and store this in the internal WaveBank instead, to keep memory use to a minimum when the flag isn't set.

flibitijibibo (changes requested):
Other than where the names are stored this works for me!

flibitijibibo (approved):

Issue #259, opened 2021-07-12
CMake FAudio config files do not respect cmake syntax


p01arst0rm:
Recently discovered this issue looking into problems with
users trying to compile with FAudio on the rpcs3 project.

this issue relates to the cmake config files. for example, the config
file is named `FAudio-config.cmake`. this is incorrect formatting for cmake,
as it is expecting a file in the format of `NameOfLibrary.cmake` or
`librarynamenocaps-config.cmake`. the upstream needs to make their
build export the file(s) either as FAudio.cmake or faudio-config.cmake

see also: https://github.com/RPCS3/rpcs3/pull/10536#pullrequestreview-704492917

flibitijibibo:
Will let you handle this since you have a test case - once you have a fix it'll go in after a quick sanity check.

p01arst0rm:
> Will let you handle this since you have a test case - once you have a fix it'll go in after a quick sanity check.

theres a couple of ways this can be fixed though. the easiest would just be changing the `project(FAudio C)` tag to `project(faudio C)`, or i could edit the config file definitions to stop using the project name to control the file name.

flibitijibibo:
Changing the project name might cause some breakages, between those two the latter sounds safer.

p01arst0rm:
alright, ill knock up a PR :)

PR #261, opened 2021-07-13
fixed cmake config name

p01arst0rm:
Fixes https://github.com/FNA-XNA/FAudio/issues/259

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -479,11 +479,11 @@ install(
 include(CMakePackageConfigHelpers)
 configure_package_config_file(
 	cmake/config.cmake.in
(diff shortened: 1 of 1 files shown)


flibitijibibo:
CCing @aeikum in case this breaks something important, looks good to me otherwise!

aeikum:
This would only affect projects which use cmake for their build and depend on FAudio, right? We don't do that in Proton (we discover it with pkg-config), so no objection from me.

flibitijibibo:
Oh, duh, totally forgot about that - thanks for checking it out anyway! I think the only product affected is Skullgirls/Indivisible then, and I can just bother the maintainer about that...

Issue #225, opened 2020-10-09
Rez Infinite asserts in FAudioFXReverb_LockForProcess()


mrpippy:
Rez Infinite from Steam asserts in `FAudioFXReverb_LockForProcess()` soon after startup, it looks like the issue is that input and output channels are both 5.1. I'm not sure if the XAudio2 reverb APO actually supports this (despite claiming not to), or if the channel counts just aren't correct. 

Log messages: https://gist.github.com/mrpippy/3edad985fd515c2d9d71e4cc8cac1261

flibitijibibo:
First guess is that we may need to return an appropriate HRESULT to get working behavior. Otherwise I can only imagine there being a path where the input/output channels match... seems unlikely though, especially 5.1->5.1 which would be _really_ expensive to process.

flibitijibibo:
My mistake... looks like we already do, but it's a pretty sloppy block: https://github.com/FNA-XNA/FAudio/blob/890f44e4f74f474524a933e0aca8962c4735e183/src/FAudio.c#L1405

I suppose we'll have to figure out how to handle this non-destructively (assuming that 5.1->5.1 path doesn't exist... man I hope it doesn't).

flibitijibibo:
Looks like the source to XAudio2.7 matches the documentation, so the above block is indeed the real issue:

https://github.com/Microsoft/DirectXTK/issues/170#issuecomment-479617599

mrpippy:
I made a test app that makes the same calls as the game, and the `SetEffectChain()` call that fails with FAudio succeeds on Windows. I guess that means it actually supports 5.1->5.1?

https://gist.github.com/mrpippy/f2ddbf57bed572c0f056612dbdf8f98f

The game later crashes when it tries to call `SetEffectParameters()` on the submix voice, but the effect chain never got set.

flibitijibibo:
Odd, it produces working sound as well? I wonder if there's a path for matching channels.

In any case this moves the issue to reverb, which is less scary but does involve messing with all the arrays to support 6<->6 processing.

mrpippy:
Yeah, the game produces working sound on Windows (just a laptop with 2 channel output)

flibitijibibo:
Some quick shortcut links...

- [Where we check the input/output format](https://github.com/FNA-XNA/FAudio/blob/master/src/FAudioFX_reverb.c#L1231)
[...]

PR #262, opened 2021-07-15
Support 5p1 -> 5p1 reverb

aeikum:
This is required by Rez Infinite, see #225. I'm pretty sure I got all of the places where channel counts are used, but please double-check, and also look over my usage of the internal reverb API.

As always, feel free to fixup any style issues in this commit.

### DIFF
--- a/src/FAudioFX_reverb.c
+++ b/src/FAudioFX_reverb.c
@@ -477,13 +477,66 @@ static float APF_OUT_DELAYS[REVERB_COUNT_APF_OUT] =
 	7.73f
 };
 
-static const float STEREO_SPREAD[4] =
+typedef enum FAudio_ChannelPositionFlags
 {
-	0.0f,
-	0.5216f,
-	0.0f,
-	0.5216f
-};
+	Position_Left = 0x1,
+	Position_Right = 0x2,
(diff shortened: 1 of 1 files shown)


aeikum:
Force-pushed a fix to averaging the input sample (used to divide by 6, now divides by 5).

flibitijibibo:
I could be misremembering but I believe DspReverb_SetParameters also has an array that assumes a max of 4 channels rather than 5, I think all it needs is an additional entry but it's been a long time since I've looked at this.

aeikum:
Good catch. That also showed me we were accessing past the end of STEREO_SPREAD. I've redone the left/right/rear logic to be based on speaker position flags. I tested how the center channel is treated on Windows with both 2.7 and 2.8, and it seems broken. It seems to behave the same as the left speaker. Instead in this code, I just averaged the left and right values for the center speaker. We already sound significantly different from Windows for the left/right parameters, so if we want to match Windows more closely, we can look into that separately.

flibitijibibo:
That's pretty wild... but yeah we're a million miles off of the Princeton sound, so I'd rather be correct than accurate in this case.

Thanks for checking this out, this looks good!

PR #265, opened 2021-09-03
Calculate Block Alignment if one wasn't specified.

alesliehughes:
Devil May Cry HD Collection creates a Voice, where the source nBlockAlign is 0.

From here
https://docs.microsoft.com/en-us/windows/win32/api/mmreg/ns-mmreg-waveformatex

We should be able to calculate the value.


alesliehughes:
This might not be right, closing until I have time to confirm

PR #266, opened 2021-09-14
For [X]WMA, send decoded packet size to the decoder

aeikum:
This patch sucks a little bit. There doesn't seem to be any standard way
to communicate the DPDS data to decoders via gstreamer.

GstAudioClippingMeta comes close, but it communicates how many samples
to trim from the resulting buffer, not how many samples we expect to be
decoded. Another option may be some form of seeking (this is how ffmpeg
consumes it when reading directly from a file), but I couldn't find a
purely informational way to communicate a seek table that wouldn't
result in some functional change.

So the best I could come up with is a custom GstMeta, which only one
decoder, Proton's media-converter, will actually consume. Not great, but
it shouldn't interfere with any other decoder.


aeikum:
Cc @rbernon , we'll need to think about this for the MF-based codepath, if we're switching to that at some point down the line.

aeikum:
Harr, I forgot to fix code style. Will update the branch shortly.

rbernon:
I'm not sure to completely understand why it is needed, is this just so that Proton transcoder can generate silence of the correct size or does it mean that normal `WMA` decoding (without Proton transcoder) doesn't work either?

aeikum:
Yes, it's only for the media-converter. FAudio uses the expected packet length to determine when it is done with a buffer. If our media-converter returns too many samples for a given packet, then FAudio can quit processing a buffer early, which means the packet stream that the media-converter receives is corrupted.

rbernon:
Is this happening across buffers or can we just push the whole input buffer at once to gstreamer / MF when a new buffer is started?

I believe I did that at some point for the MF implementation, then I changed it to only push one encoded block at a time but I think that was mostly to be nicer, I think providing the whole buffer worked as well.

aeikum:
Hmm, that's a good idea. I'll give it a look and see if I run into any problems with that approach.

aeikum:
@rbernon I've implemented that, but unfortunately performance is pretty terrible. In some cases Banished submits xaudio2 buffers that decode to over 6 seconds of audio (dpds in samples = 274432 at 44.1 kHz). Even using native gst-libav with logging disabled there are visible hitches on my fairly beefy machine. So I'm not sure if this is workable.

aeikum:
I'm experimenting with only pulling the requested amount of data from gstreamer instead of the entire decoded buffer. Looks promising.

aeikum:
Actually it's still the same problem, the decoded data just gets queued because we're in push mode. Argh.

PR #276, opened 2021-12-15
Set thread names for the Win32 platform

mrpippy:
Set thread names for the Win32 platform.

I've tested this in Wine, but not on actual Windows.

### DIFF
--- a/src/FAudio_platform_win32.c
+++ b/src/FAudio_platform_win32.c
@@ -70,6 +70,44 @@ void FAudio_Log(char const *msg)
 	OutputDebugStringA(msg);
 }
 
+static HMODULE kernelbase = NULL;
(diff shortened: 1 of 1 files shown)


flibitijibibo (approved):
Looks right to me! I don't think we need to try and implement the hacky pre-Win10 thread naming scheme so this should be fine.

Will merge when this gets a +1 from another platform_win32 maintainer, just in case.

rbernon:
I think loading `SetThreadDescription` could go in `FAudio_PlatformInit` instead of requiring a specific InitOnce, and if we take a kernelbase module ref we should probably release it in `FAudio_PlatformQuit`. Otherwise looks good.

mrpippy:
Thanks, good point, I didn't realize only one mastering voice was allowed at a time. I pushed a new version with that change and some other parts tightened up.

rbernon (approved):
LGTM

PR #278, opened 2021-12-20
Return a PCM format by default in Win32 GetDeviceDetails

rbernon:
Far Cry 4 expects it, and testing shows that it's apparently the case on Windows (https://testbot.winehq.org/JobDetails.pl?Key=104269) even though WASAPI default mix format is IEEE_FLOAT.


flibitijibibo:
Ah, pretty similar to this guy then:

https://github.com/FNA-XNA/FAudio/commit/bf69b4000b85a1fc2e24b1a112be5beef00b3e79

rbernon:
Ah yes indeed, I didn't know about it.

PR #290, opened 2022-07-01
Fix the 3-call-per-update limit on SetEffectParameters

TylerGlaiel:
treat m_uCurrentParametersIndex as a "changed" flag instead of a ring buffer index (0 = unchanged, 1 = changed), 
treat m_pCurrentParamers and m_pCurrentParametersInternal as a front and back buffer
on SetEffectParameters it writes to the back buffer and sets that flag to true
on BeginProcess it swaps those 2 if m_uCurrentParametersIndex was true, and sets the flag to false

(I would not merge this to main yet)


flibitijibibo:
Will check to see how much we can screw around with the pointers... not too hard, just have to crank up the wooden motor on the VS2010 environment that has the Jun2010 DXSDK samples installed.

TylerGlaiel:
this change should also supercede the need to call BeginProcess even if an effect is disabled (https://github.com/FNA-XNA/FAudio/commit/93a9905093b4839e195b3f3d9cfb3fe70da6f7de)

flibitijibibo:
Ran some tests against XAudio2_7 with the XAudio2CustomAPO sample and can confirm that it does indeed use the full ring buffer split into three parts, and with clever timing you can corrupt an effect if it doesn't BeginProcess before doing other processing. Not as optimal as double buffering like this but I'm sure someone from the late 2000s has an excuse ¯\_(ツ)_/¯

(inb4 a major game with XAPOs like FFXIV has this bug)

TylerGlaiel:
ah dang.... this fix was working so well internally for my purposes. does it also ignore the updates if you call SetEffectParameters 3 times in a row before BeginProcess gets called?

flibitijibibo:
No, but it is guaranteed to call it at least once: The public side sets `update = 1` and in the mixer thread (the private side) it calls the _actual_ SetParameters, so it will still always update whether it's enabled or not, it just uses 3 stages instead of just 2. To emphasize: FAudio_SetEffectParameters is _not_ the same as FAPOBase_SetParameters; there is a wall between the two to reduce the number of times the effect has to update itself.

TylerGlaiel:
ok I think I get it now, I had thought the 3-call per update limit was irrelevant to whether or not the effect was disabled (and disabling just increased the time window where you could do 3 calls to SetEffectParameters), but it was actually *entirely* because it was skipping the BeginProcess calls when it was disabled that the internal SetParameters could get called 3x without getting synced

PR #301, opened 2023-01-06
Purge and clear notifcations on ShutDown

alesliehughes:
IXACT4Engine::ShutDown states that it Flushes and unregister's all notifications.


flibitijibibo:
Just to be sure, IXACT4 or IXACT3 (or 2..?).

Other than that lgtm.

alesliehughes:
It should be IXACT3.

PR #313, opened 2023-05-04
Support for QOA decoding

thatcosmonaut:
Added an API for QOA decoding - let me know if the structure looks OK or if you would want some changes.

Reference API is here: https://github.com/phoboslab/qoa

I modified the reference to behave more like stb_vorbis where the handle maintains some state instead of having to update and pass in state yourself on every decode request, and i made the handle opaque so that we don't have to marshal on every API call.

My benchmarks show a 4x speedup over vorbis decoding on equivalent audio data.


flibitijibibo:
One way to simplify the build slightly is to implement XNA_Song with this decoder - in addition to having sample code it will also allow us to turn the decoder into a single-file header, which we can note is a third-party file and this may not follow the conventions that FAudio follows (stdint is the one that stood out, at least to me).

Aside from that it's the usual tabbing/spacing stuff, nothing major AFAICT.

(Also a reminder for myself: I need to add .qoi and .qoa as added extensions in the ContentReaders)

thatcosmonaut:
Sounds good to me. How do we want to handle the fact that XNA_Song could use vorbis or qoa?

flibitijibibo:
I believe stb knows to validate the header, so we can attempt that first and fall back to qoa after that.

visualc-gdk/FAudio.vcxproj
flibitijibibo, line 1: We should be able to undo this file change now

csharp/FAudio.cs
```diff
@@ -2385,4 +2385,34 @@ int num_samples
 	);
 
 	#endregion
+
+
```
flibitijibibo, line 2389: Looks like some double space action going on in this file...

src/XNA_Song.c
```diff
@@ -88,29 +90,55 @@ static FAudioMasteringVoice *songMaster = NULL;
...
+			(float*) songCache,
+			activeVorbisSongInfo.sample_rate * activeVorbisSongInfo.channels
+		);
+		buffer.AudioBytes = decoded * activeVorbisSongInfo.channels * sizeof(float);
+		buffer.Flags = (decoded < activeVorbisSongInfo.sample_rate) ?
+			FAUDIO_END_OF_STREAM :
+			0;
+	}
+	else
+	{
```
flibitijibibo, line 125: Might be good to assert that the qoa file is valid in these else branches

src/qoa_decoder.h
```diff
@@ -0,0 +1,524 @@
...
+	if (data->qoa.channels == 0 || data->qoa.channels == 0 || data->qoa.samplerate == 0) {
+		return 0;
+	}
+
+	return 8;
+}
+
+qoa *qoa_open_from_memory(unsigned char *bytes, unsigned int size, int free_on_close)
+{
+	qoa_data *data = (qoa_data*) FAudio_malloc(sizeof(qoa_data));
```
flibitijibibo, line 365: Similar to stb it might be easier to macro these stdlib functions ahead of the include, so we're as close to upstream as possible

flibitijibibo (changes requested):
A couple broad changes, functionally it looks okay!

thatcosmonaut:
Feedback addressed!

flibitijibibo (approved):

PR #314, opened 2023-05-12
Fix incorrect string marshaling

thatcosmonaut:


kg (approved):

flibitijibibo:
This may need to be LPCStr, or whatever I did for Vorbis, as Utf8 isn't valid with Framework.

PR #316, opened 2023-06-06
ParseSoundBank: Use Offset where possible.

alesliehughes:
This is one step in supporting older xactengine dlls.


src/FACT_internal.c
```diff
@@ -2817,7 +2817,8 @@ uint32_t FACT_INTERNAL_ParseSoundBank(
 	cur = 0;
 
 	/* Simple Cue data */
-	FAudio_assert(cueSimpleCount == 0 || (ptr - start) == cueSimpleOffset);
+	if (cueSimpleCount > 0)
+		ptr = start + cueSimpleOffset;
```
flibitijibibo, line 2821: We could probably indent the whole block here, even though it adds to the diff size

src/FACT_internal.c
```diff
@@ -2831,7 +2832,8 @@ uint32_t FACT_INTERNAL_ParseSoundBank(
 	}
 
 	/* Complex Cue data */
-	FAudio_assert(cueComplexCount == 0 || (ptr - start) == cueComplexOffset);
+	if (cueComplexCount > 0)
+		ptr = start + cueComplexOffset;
```
flibitijibibo, line 2836: Same here

flibitijibibo (commented):
Couple minor visual nits, otherwise is okay - I do wonder if having a "strict" mode would still be helpful; the asserts did find some pretty subtle parsing errors way back in the early days.

Issue #329, opened 2024-01-21
Playing VVVVVV on the Wii gives just noise


mardy:
Hi all! I'm trying to port the [VVVVVV](https://github.com/TerryCavanagh/VVVVVV) game on the Nintendo Wii, and instead of music I hear just noise. It sounds like a problem due to the processor endianness (the Wii has a big endian PowerPC). The SDL audio tests do work if I prepare an AUDIO_F32 wav file, so I tend to thing that the problem cannot be on the SDL side.

VVVVVV keeps its music in ogg format, if that matters.

Has FAudio been tested on big endian architectures? It may also be that the bug is in the game's side, because I had to modify it a little to run on the Wii.

flibitijibibo:
I think PowerPC Macs have been tested, but I'm sure FAudio_internal.c and possibly FAudio_internal_simd.c could use some more auditing in that area.

mardy:
Hi again! So, some more information: it's (most likely) not a problem in the vorbis backend, because wav files are affected too.

Then, I did this silly change in the `FAudio_INTERNAL_MixCallback` which feeds audio to SDL, and it makes the problem disappear:
```diff
diff --git a/src/FAudio_platform_sdl2.c b/src/FAudio_platform_sdl2.c
index e09f7b8..820212e 100644
--- a/src/FAudio_platform_sdl2.c
+++ b/src/FAudio_platform_sdl2.c
@@ -47,6 +47,14 @@ static void FAudio_INTERNAL_MixCallback(void *userdata, Uint8 *stream, int len)
                        audio,
                        (float*) stream
                );
+               for (int i = 0; i < len / 4; i++) {
+                       uint32_t *word = (uint32_t*)&stream[i * 4];
+                       *word =
+                               ((0xff000000 & *word) >> 24) |
+                               ((0x00ff0000 & *word) >> 8) |
+                               ((0x0000ff00 & *word) << 8) |
+                               ((0x000000ff & *word) << 24);
+               }
        }
 }
 
```

Of course, this is not a solution, but I think it proves that this is an endianness issue in FAudio itself?

flibitijibibo:
Seems that way - definitely happy to merge in changed to the mixer to ensure correct endian behavior for all input formats. For VVV specifically I would guess that this could be fixed in the FAudio_Decode_* functions involving PCM16.

mardy:
Looks like the fix is easier than expected :-D

PR #330, opened 2024-01-22
SDL: set the playback format with native endianness

mardy:
This fixes sound playback on big-endian machines.

Fixes: #329


flibitijibibo:
Nothing like a good endian fix

image

Issue #326, opened 2023-12-13
Document graph dependency issues for DestroyVoice


flibitijibibo:
Example: Destroying a submix that a voice sends data to will crash, this is avoidable in the library but it's a whole lot easier to avoid on the application side, so let's clarify this behavior.

PR #331, opened 2024-02-27
Don't destroy voice when it is output to other voices

gofman:
Cosmoteer (Steam game id 799600) sometimes destroys a submix voice which is an output for another voice. That leads to use after free access in audio client thread performing mixing which leads to obscure random crashes.

MS docs [1] suggest that IXAudio2Voice::DestroyVoice fails in such a case ("If any other voice is currently sending audio to this voice, the method fails.").

I made a test on top of Wine xaudio2 tests which indirectly confirms that (attaching a patch). I don't immediately see a direct way to test if DestroyVoice actually failed or not as it doesn't return any status. However, I found that after successful voice destruction IXAudio2SubmixVoice_GetVoiceDetails() for that voice crashes on Windows and using this fact to probe for actual voice destruction (the first one in the test doesn't crash as, the second one guarded with if (0) crashes for me on Win10).

1. https://learn.microsoft.com/en-us/windows/win32/api/xaudio2/nf-xaudio2-ixaudio2voice-destroyvoice

[0001-xaudio2-tests-Test-destroying-a-voice-which-is-an-ou.txt](https://github.com/FNA-XNA/FAudio/files/14413256/0001-xaudio2-tests-Test-destroying-a-voice-which-is-an-ou.txt)

### DIFF
--- a/include/FAudio.h
+++ b/include/FAudio.h
@@ -1064,6 +1064,11 @@ FAUDIOAPI void FAudioVoice_GetOutputMatrix(
 /* Removes this voice from the audio graph and frees memory. */
 FAUDIOAPI void FAudioVoice_DestroyVoice(FAudioVoice *voice);
 
+/*
+ * Returns S_OK on success and E_FAIL if voice could not be destroyed (e. g., because it is in use).
+ */
+FAUDIOAPI uint32_t FAudioVoice_DestroyVoiceSafeEXT(FAudioVoice *voice);
+
 /* FAudioSourceVoice Interface */
 
 /* Starts processing for a source voice.
--- a/src/FAudio.c
+++ b/src/FAudio.c
@@ -2265,10 +2265,70 @@ void FAudioVoice_GetOutputMatrix(
 	LOG_API_EXIT(voice->audio)
 }
 
-void FAudioVoice_DestroyVoice(FAudioVoice *voice)
+static uint32_t check_for_sends_to_voice(FAudioVoice *voice)
 {
+	FAudio *audio = voice->audio;
+	uint32_t ret = 0;
+	FAudioSourceVoice *source;
+	FAudioSubmixVoice *submix;
+	LinkedList *list;
 	uint32_t i;
+
+	FAudio_PlatformLockMutex(audio->sourceLock);
+	list = audio->sources;
+	while (list != NULL)
+	{
+		source = (FAudioSourceVoice*) list->entry;
+		for (i = 0; i < source->sends.SendCount; i += 1)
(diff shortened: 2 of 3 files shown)


include/FAudio.h
```diff
@@ -1062,7 +1062,7 @@ FAUDIOAPI void FAudioVoice_GetOutputMatrix(
 );
 
 /* Removes this voice from the audio graph and frees memory. */
-FAUDIOAPI void FAudioVoice_DestroyVoice(FAudioVoice *voice);
+FAUDIOAPI uint32_t FAudioVoice_DestroyVoice(FAudioVoice *voice);
```
flibitijibibo, line 1065: I think this counts as an ABI break, so we may have to export a new function that includes the return value :(

src/FAudio.c
```diff
@@ -2265,11 +2265,69 @@ void FAudioVoice_GetOutputMatrix(
...
 
+	if (!check_for_sends_to_voice(voice))
+	{
+		LOG_ERROR(
+			voice->audio,
+			"Voice %p is an output for other voice(s)",
+			voice
+		)
+		LOG_API_EXIT(voice->audio)
+		return 0;
```
flibitijibibo, line 2328: Should this be E_FAIL and the one below be 0?

flibitijibibo (commented):
At a high level this makes sense to me, we may have to tweak things slightly though - I think this would actually fix #326 as well, so bringing in @thatcosmonaut for review.

flibitijibibo:
Actually it turns out the signature is correct...

https://learn.microsoft.com/en-us/windows/win32/api/xaudio2/nf-xaudio2-ixaudio2voice-destroyvoice

... but honestly I'd be up for exporting a new function since being able to see this failure seems important. DestroyVoiceSafeEXT maybe?

gofman:
Yes, sure. I did it this way because technically it doesn't break the ABI, if old client uses void version and doesn't check value it is not broken with newer version returning the value. But it is looks cleaner to have a different function. I think both old and new one will check if the voice is in use and the difference will be only in the presence return value. Or do you want the old version work like before?

flibitijibibo:
What we can probably do is have this in the header...

```
FAUDIOAPI void FAudioVoice_DestroyVoice(...);

FAUDIOAPI uint32_t FAudioVoice_DestroyVoiceSafeEXT(...);
```

... and then we'll rename the implementation we have now to DestroyVoiceSafeEXT, and DestroyVoice can call DestroyVoiceSafeEXT and discard the return value entirely. It's a bit roundabout, but we've done similar things with other extensions like voice filters and it's been easy enough to keep stable!

flibitijibibo (approved):

PR #333, opened 2024-03-06
build: fix building with MSVC

rkitover:
Increase C standard level to 11 for alignas(x) instead of relying on gcc/clang-specific __attribute__((aligned(x))).

Link system DLLs on Windows by name instead of as an -l linker flag, cmake automatically produces the correct linker invocation on both MINGW and MSVC.

Include <windows.h> with WIN32_LEAN_AND_MEAN defined instead of including <windef.h> and <winbase.h>, because this breaks the Windows header architecture detection on MSVC.

Define the GUIDs for IID_IAudioClient, IID_IAudioRenderClient, IID_IMMDeviceEnumerator and CLSID_MMDeviceEnumerator for PLATFORM_WIN32 taken from the SDK headers, because MSVC cannot find the linkage for them.


flibitijibibo:
All looks good except for the language version bump - does __declspec(align) work with newer VS versions? We should be able to stay on the old version if so.

rkitover:
How's this, I added an `ALIGN(type, boundary)` macro.

rkitover:
Can you rerun the CI please, I fixed the MINGW fail.

PR #334, opened 2024-03-08
build: fix __attribute__((aligned(x))) on gcc/clang

rkitover:
This is a followup on 81d5a24 (build: fix building with MSVC,
2024-03-07).

Fix the missing trailing _ in __attribute__ for
__attribute((aligned(x))).

### DIFF
--- a/src/FAudio_internal.h
+++ b/src/FAudio_internal.h
@@ -211,8 +211,8 @@ extern void FAudio_Log(char const *msg);
 #endif
 
 /* Alignment macro for gcc/clang/msvc */
-#if defined(__GNUC__) || defined(__clang__)
-#define ALIGN(type, boundary) type __attribute_((aligned(boundary)))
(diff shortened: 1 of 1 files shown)


flibitijibibo:
This isn't safe AFAIK, but we've been using this for some time now... I'll forward this to the Apple peoples on the FNA Discord to see what might be up.

flibitijibibo:
That would probably be okay, though we're still trying to figure out if there's a way to make this work correctly on M1. Will make the call this weekend!

kiddkaffeine:
@rkitover, did this exist prior to the other change, or is it a regression?  How does it manifest?  Crash?  Corrupted audio?  Is this specific to M1?  Wouldn't it affect iOS/tvOS as well?

rkitover:
@flibitijibibo I'm sorry, but this was all caused because I made a typo, I wrote `__attribute_` instead of `__attribute__`, this works correctly now.

PR #338, opened 2024-04-07
[windows] List all audio devices for Windows

Steelskin:
Previously, FAudio with `PLATFORM_WIN32` would only list the default audio device. This reworks the Faudio `PLATFORM_WIN32` code to enable access to every audio device on the system. In order to preserve the existing behavior (and Xaudio2 compabitility), the audio devices are re-ordered so the first device is always the default one.

In addition, this properly populates the `DisplayName` field in `FAudioDeviceDetails`. Previously, it was set to the device GUID.

Test: Local build with visualboyadvance-m

### DIFF
--- a/src/FAudio_platform_win32.c
+++ b/src/FAudio_platform_win32.c
@@ -41,6 +41,7 @@
 #include <initguid.h>
 #include <audioclient.h>
 #include <mmdeviceapi.h>
+#include <Functiondiscoverykeys_devpkey.h>
 
 DEFINE_GUID(CLSID_CWMADecMediaObject, 0x2eeb4adf, 0x4578, 0x4d10, 0xbc, 0xa7, 0xbb, 0x95, 0x5f, 0x56, 0x32, 0x0a);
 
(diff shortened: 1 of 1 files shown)


Steelskin:
Friendly ping. Do you have any concern or issue with this PR?

flibitijibibo:
I don't personally, but this is code meant for Wine so I was hoping it would get reviewed over there... we may need to submit this as a patch to Wine first? Here's their local copy:

https://gitlab.winehq.org/wine/wine/-/tree/master/libs/faudio?ref_type=heads

src/FAudio_platform_win32.c
```diff
@@ -41,6 +41,7 @@
 #include <initguid.h>
 #include <audioclient.h>
 #include <mmdeviceapi.h>
+#include <Functiondiscoverykeys_devpkey.h>
```
rbernon, line 44: Could you use `devpkey.h` and then, `DEVPKEY_Device_FriendlyName` instead? Fwiw, `Functiondiscoverykeys_devpkey.h` doesn't exist in Wine.

src/FAudio_platform_win32.c
```diff
@@ -202,6 +203,128 @@ static DWORD WINAPI FAudio_AudioClientThread(void *user)
...
+	CoTaskMemFree(default_guid);
+	return 0;
+}
+
+/* Open `device`, corresponding to `deviceIndex`. `deviceIndex` 0 always
+ * corresponds to the default device. XAudio reorders the devices so that the
+ * default device is always at index 0, so we mimick this behavior here by
+ * swapping the devices at indexes 0 and `defaultDeviceIndex`.
+ */
+static DWORD WINAPI FAudio_OpenDevice(uint32_t deviceIndex, IMMDevice **device)
```
rbernon, line 284: Should probably return `HRESULT`, and `WINAPI` isn't needed so it should be dropped.

src/FAudio_platform_win32.c
```diff
@@ -202,6 +203,128 @@ static DWORD WINAPI FAudio_AudioClientThread(void *user)
 	return 0;
 }
 
+/* Returns the device index for the default device.
+ * Returns 0 on failure. This will prevent the rest of the FAudio code from 
+ * re-ordering the devices. */
+static uint32_t FAudio_DefaultDeviceIndex(IMMDeviceCollection *device_collection)
```
rbernon, line 209: This returns 0 on any error, is this intended? Maybe this should instead return it to the caller?

src/FAudio_platform_win32.c
Steelskin: Done.

src/FAudio_platform_win32.c
Steelskin: Done.

src/FAudio_platform_win32.c
Steelskin: I reworked this to return an `HRESULT` and take `defaultDeviceIndex` as parameter here. It doesn't really change much as setting `defaultDeviceIndex` to 0 just does not change the ordering. We just end up with the same error down the line so might as well return the error code here.

Steelskin (commented):
Thanks for getting back to me! I addressed your comments.

rbernon (commented):
Looks good to me, with some slightly inconsistent style issues, though I don't know how picky I'm expected to be. I think squashing the changes together would also be better.

flibitijibibo:
We can be pretty picky about style, but that can always come through as style fix commits - will get this in now since it's been waiting so patiently!

PR #339, opened 2024-04-07
[windows] Exit cleanly on audio device disconnect

Steelskin:
Previously, we would crash when the audio device was disconnected.


Steelskin:
@rbernon I have this one waiting on a review too. PTAL.

rbernon (approved):

PR #342, opened 2024-04-21
Fix deadlock in voice callbacks

xtsm:
Second attempt. Previous here: https://github.com/FNA-XNA/FAudio/pull/337

Now that we release buffer lock before executing callbacks, some code from another thread can get scheduled between mutex unlock and callback call and observe empty buffer queue right before callback puts a new buffer there. XNA_Song determines the end of the song by simply checking if the queue is empty, which may lead to random playback breaks.

Not sure how to properly check for the song end but the buffer decoding function also zeroes out total samples counter when it finishes a EOS-flagged buffer so checking if that counter is zero should cut it... probably?

### DIFF
--- a/src/FAudio_internal.c
+++ b/src/FAudio_internal.c
@@ -383,6 +383,12 @@ static void FAudio_INTERNAL_DecodeBuffers(
 			if (	voice->src.callback != NULL &&
 				voice->src.callback->OnBufferStart != NULL	)
 			{
+				FAudio_PlatformUnlockMutex(voice->src.bufferLock);
+				LOG_MUTEX_UNLOCK(voice->audio, voice->src.bufferLock)
+
+				FAudio_PlatformUnlockMutex(voice->sendLock);
+				LOG_MUTEX_UNLOCK(voice->audio, voice->sendLock)
+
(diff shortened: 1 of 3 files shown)


xtsm:
@Romans-I-XVI could you please check if the problem you reported earlier (https://github.com/FNA-XNA/FAudio/pull/337#issuecomment-2049854555) persists in this branch?

Romans-I-XVI:
> @Romans-I-XVI could you please check if the problem you reported earlier ([#337 (comment)](https://github.com/FNA-XNA/FAudio/pull/337#issuecomment-2049854555)) persists in this branch?

Just tested it out, this fixes the issue I was experiencing.

flibitijibibo:
Sounds good - we're pretty close to the tag date so I'll merge this in right after 24.05 is out!

xtsm:
@flibitijibibo no tag yet? i thought FNA 24.05 already got released

flibitijibibo:
Fell behind due to SDL_GPU and some other things, will do a full review later today to be extra sure that this is solid.

flibitijibibo (approved):
We should probably isolate the callbacks in such a way that we won't have to copypaste these unlock/lock pairs so much... but that's for another day.

Issue #336, opened 2024-03-10
Moorhuhn Director's Cut crashes inside FACTSoundBank_Prepare


gverm:
I hope it's okay to report this here:

A user has reported that Moorhuhn Director's Cut crashes inside xactengine. Sometimes it happens immediately when selecting the level, sometimes you get a few seconds of audio/gameplay before it crashes.

Related bugzilla report: https://bugs.winehq.org/show_bug.cgi?id=55939

Backtrace is as following:

```
Backtrace:
=>0 0x7fcc7a1e FACTSoundBank_Prepare+0x19e(pSoundBank=<is not available>, nCueIndex=<is not available>, dwFlags=<is not available>, timeOffset=<is not available>, ppCue=<is not available>) [/home/gverm/Desktop/32Bit/../wine/libs/faudio/src/FACT.c:1242] in xactengine2_9 (0x00659e80)
  1 0x7fcc8bb7 FACTSoundBank_Play+0x47(pSoundBank=00357668, nCueIndex=0xffff, dwFlags=0, timeOffset=0, ppCue=00000000) [/home/gverm/Desktop/32Bit/../wine/libs/faudio/src/FACT.c:1308] in xactengine2_9 (0x00659eac)
  2 0x7fcc406b IXACT3SoundBankImpl_Play+0x12b(iface=01766038, nCueIndex=0xffff, dwFlags=0, timeOffset=0, ppCue=00000000) [/home/gverm/Desktop/32Bit/../wine/dlls/xactengine3_7/xact_dll.c:504] in xactengine2_9 (0x00659eec)
  3 0x00401866 in mhdc (+0x1866) (0x00659fcc)
  4 0x00440301 in mhdc (+0x40301) (0x02428c18)
  5 0x01860148 (0x004fe06c)
  6 0x0043b910 in mhdc (+0x3b910) (0x0043b8d0)
  7 0x0004d8e8 (0xe8f18b56)
0x7fcc7a1e FACTSoundBank_Prepare+0x19e [/home/gverm/Desktop/32Bit/../wine/libs/faudio/src/FACT.c:1242] in xactengine2_9: cmpb $0x03, (%eax)
1242			if ((*ppCue)->variation->flags == 3)
```

Full log: [mhdc.txt](https://github.com/FNA-XNA/FAudio/files/14551817/mhdc.txt)

There is a download available at archive.org here: https://archive.org/details/moorhuhn-directors-cut

flibitijibibo:
[...]

PR #344, opened 2024-04-27
Support cues which have no variation

alesliehughes:
This will stop the crash in #336 .


src/FACT_internal.c
```diff
@@ -515,7 +515,7 @@ uint8_t FACT_INTERNAL_CreateSound(FACTCue *cue, uint16_t fadeInMS)
 		/* Sound */
 		baseSound = cue->sound;
 	}
-	else
+	else if(cue->variation)
```
flibitijibibo, line 518:
```suggestion
	else if (cue->variation)
```

flibitijibibo (approved):
lgtm, just adding a space and squashing...

PR #351, opened 2024-06-30
build: separate shared and static import targets

rkitover:
Export the FAudio::FAudio-shared target when building a shared library and the FAudio::FAudio-static target when building a static library to separate target files so that both can be installed at the same time.

Alias FAudio::FAudio to the shared library if it is installed and the static library otherwise in the Config.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,8 +64,14 @@ if(INSTALL_MINGW_DEPENDENCIES)
 	include(cmake/install_shared_libs.cmake)
 endif()
 
+set(target FAudio-shared)
+
+if(NOT BUILD_SHARED_LIBS)
+	set(target FAudio-static)
+endif()
+
 # Source lists
-add_library(FAudio
+add_library(${target}
 	# Public Headers
 	include/F3DAudio.h
 	include/FACT3D.h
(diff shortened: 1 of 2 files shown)


flibitijibibo (commented):
This change looks okay but since we're tagging in about 12 hours I'll wait until 24.07 is out - expect a merge soon!

rkitover:
Can this be merged now, or is there some legal order preventing that from happening? And are you actually forced to comply with it by any means?

flibitijibibo:
No legal barriers - I and the other maintainers are just stuck on SDL_GPU at the moment.

That said, it's still a failure on my part for not coordinating this better, so if it's okay I'd like to defer this to someone who is using this for similar purposes. @cybik, sorry to bother you but does this adversely affect the projects making use of the current CMake configurations?

cybik:
Full disclosure: I don't touch such projects often these days.

That being said: the last active project I recall using this, is using a frozen version and rarely updates. I'll reach out to them to make them aware of this change should they update FAudio; beyond that, since the *default* is `BUILD_SHARED_LIBS ON`, I'm going to assume it'll be fine. If the default was changed to *static*, legalities COULD end up being rude.

rkitover:
The main point of this is to allow a distro to do two builds with both `-DBUILD_SHARED_LIBS=ON` and `-DBUILD_SHARED_LIBS=OFF`, like MSYS2 does, and allow both to be installed simultaneously without the second build overwriting the CMake target import configuration of the first.

And when both are installed, it makes an `FAudio::FAudio-static` target available for static builds,

Currently the way I link FAudio in our project is by listing the libraries and its dependents because there is no target.

I have followed the convention here used by some other projects such as SDL2, which makes an `SDL2::SDL2-static` target available for this purpose.

In all other respects, this changes nothing.

We should really support building both targets from a single build, but that's a separate issue.

rkitover:
I thought it would be reasonable to make the `FAudio::FAudio` target point to the static libs if **ONLY** the static libs are installed on a distribution, but I can remove this if you are concerned about there being any confusion about this.

rkitover:
Actually now that I think about it, without this change, if only the static configuration is installed, then the `FAudio::FAudio` target points to the static libs as well, so that is also the same.

PR #364, opened 2024-10-31
Add C# wrapper for part of FAPO and FAPOBase APIs

kg:
Just enough to be able to write your own spectrum analyzer.

It would be nice if FAudio took ownership of the registration properties passed in, so I could make them a `ref T` argument instead of an IntPtr you have to malloc and leak.

### DIFF
--- a/csharp/FAudio.cs
+++ b/csharp/FAudio.cs
@@ -947,9 +947,9 @@ public static class FAudio
 	public delegate uint LockForProcessFunc (
 		IntPtr fapo,
 		uint InputLockedParameterCount,
-		IntPtr pInputLockedParameters,
+		ref FAPOLockForProcessBufferParameters pInputLockedParameters,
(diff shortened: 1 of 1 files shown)


kg:
I've been using this successfully for a while so it's probably good enough.

csharp/FAudio.cs
```diff
@@ -27,6 +27,7 @@
 #region Using Statements
 using System;
 using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.ComTypes;
```
flibitijibibo, line 30: Which type needed this namespace?

flibitijibibo (commented):
Overall fine minus the one using statement

csharp/FAudio.cs
kg: This is probably just VS2022 being incredibly stupid, it loves to add random using statements now

PR #370, opened 2024-12-18
FACT: Comment out the 0ms fade handler.

### LINKED ISSUE #368: Paused Cues are Stopped and Paused when a second Cue is played
Further to the [thread on Discord](https://discord.com/channels/476524251441004544/1315349488558149652), it seems that playing a Cue after previously pausing another one makes the paused cue simultaneously enter both a paused and stopped state. While I can't reproduce the expected behaviour any more due to a lack of access to XNA, the game I wrote back in 2010 depended on this behaviour to prevent players going insane from listening to the first few bars of the overworld theme repeatedly 😁 

## Versions

* FNA 24.12, d12e1b0
* MacOS Sequoia 15.1.1
* .NET 8.0.111

## Observed behaviour

* Consider two `Cue` instances, `a` and `b`
* `a.Play()`
* `a.Pause()`
* `b.Play()`
* `a.IsPaused && a.IsStopped == true`

## Expected behaviour

* `a.Play()`
* `a.Pause()`
* `b.Play()`
* `a.IsPaused && !a.IsStopped`

...then for bonus points:

* `b.Pause()`
* `a.Resume()`
* `a` starts playing from the point at which it was previously paused

## Repro steps

I made a [repro game](https://github.com/DanielJonesEB/fna-faudio-repro) that depends on Git LFS, a local checkout of FNA, and FNA libraries to be manually copied in that throws an exception if the (suspected) bug is present.

## Other possible explanations

* It's entirely possible that something changed in XACT between XNA 3.0 and XNA 4 that meant that I've forgotten to change a setting. I've been updating my game project from XNA 3.0 after having ignored it for 14 years, so there's lots of context I'll have missed. So it could be that the library is fine, but the problem is in my content.
* Could be something to do with MacOS specifically?

flibitijibibo:
There's a 1-frame timing issue where two sounds in the same category being played consecutively should still have both sounds _technically_ playing, even if the fade time is 0ms.

Should fix #368.

### DIFF
--- a/src/FACT_internal.c
+++ b/src/FACT_internal.c
@@ -940,12 +940,18 @@ void FACT_INTERNAL_DestroySound(FACTSoundInstance *sound)
 
 void FACT_INTERNAL_BeginFadeOut(FACTSoundInstance *sound, uint16_t fadeOutMS)
 {
+	/* FIXME: Even for 0ms it seems to want to be in STOPPING for one frame.
+ 	 * Should we still destroy but defer state changes to the next frame?
+   	 * -flibit
+     	 */
+#if 0
 	if (fadeOutMS == 0)
 	{
 		/* No fade? Screw it, just delete us */
 		FACT_INTERNAL_DestroySound(sound);
 		return;
 	}
+#endif
 
 	sound->fadeType = 2; /* Out */
 	sound->fadeStart = FAudio_timems();


flibitijibibo:
@DanielJonesEB does this FAudio dylib fix the issue? https://github.com/FNA-XNA/FAudio/actions/runs/12393993213/artifacts/2337691633

DanielJonesEB:
Hey @flibitijibibo! I'm just getting back up to speed after the Christmas. Is there a compiled build I can download to test it, or should I attempt to build from source?

flibitijibibo:
Rebased this patch just now, an SDL2 osx dylib should show up toward the bottom of this page: https://github.com/FNA-XNA/FAudio/actions/runs/12636518921

PR #371, opened 2025-02-07
Add FAudioFX_Collector

kg:
This is a simple FAPO that takes a user-provided sample ring buffer and sample counter and will write incoming samples into it. It averages all incoming channels so the user doesn't need to know the channel count. The sample counter increases once per input block from the mixer, which allows the user to detect whether a new block of samples has been processed since the last time they checked.
The sample counter is written using atomics so the user won't observe a torn write, and since the ring buffer is updated one float at a time it's safe for us to race against the consumer without a lock (the odds of them reading from the buffer while we're writing to it are quite low, and for most use cases it would be unobservable).

This is designed to work great as a way to feed a spectrum analyzer or realtime waveform display, where you just want a buffer with N recent samples into it independent of the mixer resolution - for an FFT you probably want a buffer size that's a power of two, for example.

Haven't been able to test this yet so this draft is mostly to solicit API design feedback etc. and give people time to think about it.

### DIFF
--- a/include/FAPO.h
+++ b/include/FAPO.h
@@ -84,6 +84,7 @@ typedef struct FAPOProcessBufferParameters
 {
 	void* pBuffer;
 	FAPOBufferFlags BufferFlags;
+	// The number of per-channel frames (samples) in the buffer. The total size of the buffer in floats is (ValidFrameCount * ChannelCount).
 	uint32_t ValidFrameCount;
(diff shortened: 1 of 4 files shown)


flibitijibibo:
Superceded by #377

kg:
Sorry, didn't realize I had created two PRs...

PR #375, opened 2025-03-16
Making sure output debug messages are \n terminated.

sherief:
Prior to this, debug output with no \n results in multiple calls to OutputDebugStringA() on Windows resulting in some very, very long lines in (for example) the Output window in Visual Studio's debugger. In Visual Studio's case, this is enough to degrade UI performance to unusable points.

### DIFF
--- a/src/FAudio_internal.c
+++ b/src/FAudio_internal.c
@@ -81,13 +81,22 @@ void FAudio_INTERNAL_debug(
 
 	/* The actual message... */
 	va_start(va, fmt);
-	FAudio_vsnprintf(
+	out += FAudio_vsnprintf(
 		out,
 		sizeof(output) - (out - output),
 		fmt,
 		va
 	);
 	va_end(va);
+
+    char* terminator = output + (1024 - 2);
+    if (out < terminator)
+    {
+        terminator = out;
+    }
+
+    *terminator++ = '\n';
+    *terminator++ = '\0';
 
 	/* Print, finally. */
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Not a huge deal but if this can be moved to the Win32 platform that'll avoid some extra work in the SDL side which does this via SDL_Log.

sherief:
totally fine by me - I was working on another PR that adds support for logging to callback (which I understand could be more intrusive and might not be suitable / approved for merge), so was wondering whether long term the right place for this would be the log function itself or the platform layers?

sherief:
https://github.com/FNA-XNA/FAudio/pull/376

if you think that's a good idea I wouldn't mind modifying the SDL backend to remove its \n insertion - would that be okay?

flibitijibibo:
The weird thing in this case is that SDL sanitizes the end so that it both adds a new line when no new line is present, but also preserves 1 or more newlines if they're already present:

Newline cutoff: https://github.com/libsdl-org/SDL/blob/main/src/SDL_log.c#L613

Newline guarantee: https://github.com/libsdl-org/SDL/blob/main/src/SDL_log.c#L645

So we may just have to rip off SDL here and snprintf in Win32.c if we want it to be consistent. (I should probably document FNA3D to mention this, we follow the same behavior there as a result...)

sherief:
okay, that makes sense - thanks for clearing this up. I think this PR should be closed without merging now.

PR #376, opened 2025-03-16
Supporting optional logging to user-provided callback.

sherief:
Some users might want to redirect this, or reconcile it with logging from other subsystems in a single sink.

### DIFF
--- a/include/FAudio.h
+++ b/include/FAudio.h
@@ -316,6 +316,8 @@ typedef struct FAudioDebugConfiguration
 	int32_t LogFileline;
 	int32_t LogFunctionName;
 	int32_t LogTiming;
+    void (*LogCallback)(const char* msg, void* userData);
+    void* LogUserData;
 } FAudioDebugConfiguration;
 
 #pragma pack(pop)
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Sadly we can't change the public structs but we can totally add something similar to FNA3D's log rerouting:

https://github.com/FNA-XNA/FNA3D/blob/master/include/FNA3D.h#L496

sherief:
I see, makes sense. Would it be acceptable if I were to add a new struct / function for this?

flibitijibibo:
Definitely- we can do FAudio_HookLogFunction(s)EXT and have it be global like FNA3D, so that we don't have to fuss with object lifetime to get logs routed properly.

sherief:
I host multiple instances in the same process and would prefer their output be separated - what do you think is the best path forward here?

flibitijibibo:
If you've got an active test case then we can accommodate that - usually I just shotgun blast debug stuff unless somebody has a specific need, so if we need to put a function pointer in the FAudio object that's fine too.

PR #377, opened 2025-04-01
Add FAudioCreateCollector to create a FAPO that captures samples into a buffer

kg:
Simple and easy to use. Hand it a ring buffer and it will fill it with mono samples (so you don't need to think about how many channels there are), and you can feed that into an FFT or something else in order to do a realtime spectrum analyzer, etc.

I previously did this with a custom callback from outside but Collector makes it much easier and eliminates the possibility of crashing/hanging the mixer thread.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,6 +100,7 @@ add_library(${target}
 	src/FAPOFX_reverb.c
 	src/FAudio.c
 	src/FAudioFX_reverb.c
+	src/FAudioFX_collector.c
 	src/FAudioFX_volumemeter.c
 	src/FAudio_internal.c
 	src/FAudio_internal_simd.c
--- a/csharp/FAudio.cs
+++ b/csharp/FAudio.cs
@@ -1071,7 +1071,12 @@ public static class FAudio
 	}
 
 	[DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl)]
-	public static extern uint FAudioCreateCollector(
+	public static extern uint FAudioCollectorGetWritePositionEXT(
+		IntPtr pApo
+	);
+
+	[DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl)]
+	public static extern uint FAudioCreateCollectorEXT(
 		out IntPtr ppApo,
 		UInt32 flags,
 		IntPtr pBuffer,
--- a/csharp/FAudio.cs
+++ b/csharp/FAudio.cs
@@ -1071,6 +1071,25 @@ public static class FAudio
 	}
 
 	[DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl)]
+	public static extern uint FAudioCreateCollector(
+		out IntPtr ppApo,
+		UInt32 flags,
+		IntPtr pBuffer,
+		UInt32 bufferLength
+	);
+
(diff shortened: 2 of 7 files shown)


flibitijibibo:
Overall looks good - let's add this as a full extension and add a text document in the extensions/ folder.

kg:
Do you think it's necessary to have some sort of mechanism to allow the user to figure out where the write position is so they can avoid processing the same data twice? I had that originally but it felt like unnecessary complication, so I stripped it out.
Another option would be to have two buffers and a flag that tells you which one is "ready", I guess.

Do we have any documentation on the extension process anywhere? I'm not sure if I need to do anything beyond documenting it.

flibitijibibo:
I think VolumeMeter has something similar but if not I wouldn't worry about it. I usually just copy another extension file when making a new one, it's usually pretty short unless there's something super finicky about how it works.

kg:
Updated with an extension description txt along with some other stuff I overlooked. Not sure I got everything.

flibitijibibo (changes requested):
Finally was able to sit down and look at this in more detail - the write position should be done with `FAudioVoice_GetEffectParameters` instead of a new export since that will avoid race conditions on the XAPO's internal variable:

```
// FAudioFX.h
typedef struct FAudioFXCollectorParametersEXT
{
    uint32_t writeOffset;
} FAudioFXCollectorParametersEXT;

// FAudioFX_collector.c
void FAudioFXCollector_GetParameters(
	FAudioFXCollector *fapo,
	FAudioFXCollectorParametersEXT *pParameters,
	uint32_t ParameterByteSize
) {
	FAudioFXCollectorParametersEXT *levels = (FAudioFXCollectorParametersEXT*)
		fapo->base.m_pCurrentParameters;
	FAudio_assert(ParameterByteSize == sizeof(FAudioFXCollectorParametersEXT));

	pParameters->writeOffset = fapo->writeOffset;
}
```

This _may_ also be a better place to fill a user buffer, for basically the same reason - I think this is why VolumeMeter has its own ring buffer internally.

kg:
> Finally was able to sit down and look at this in more detail - the write position should be done with `FAudioVoice_GetEffectParameters` instead of a new export since that will avoid race conditions on the XAPO's internal variable:
> 
> ```
> // FAudioFX.h
> typedef struct FAudioFXCollectorParametersEXT
> {
>     uint32_t writeOffset;
> } FAudioFXCollectorParametersEXT;
> 
> // FAudioFX_collector.c
> void FAudioFXCollector_GetParameters(
> 	FAudioFXCollector *fapo,
> 	FAudioFXCollectorParametersEXT *pParameters,
> 	uint32_t ParameterByteSize
> ) {
> 	FAudioFXCollectorParametersEXT *levels = (FAudioFXCollectorParametersEXT*)
> 		fapo->base.m_pCurrentParameters;
> 	FAudio_assert(ParameterByteSize == sizeof(FAudioFXCollectorParametersEXT));
> 
> 	pParameters->writeOffset = fapo->writeOffset;
> }
> ```
> 
> This _may_ also be a better place to fill a user buffer, for basically the same reason - I think this is why VolumeMeter has its own ring buffer internally.

There shouldn't be a race condition with the existing way. I'm not sure I see the advantage of this change, is it just for consistency?

flibitijibibo (approved):

PR #382, opened 2025-05-21
Do not reference count audio engine from voices

gofman:
My patch c00b01caa6e3f0dce09af6676ff55936e97b451c ("Don't destroy voice when it is output to other voices") technically caused a regression with some games under Wine reported here: https://bugs.winehq.org/show_bug.cgi?id=58253

The blamed  commit still looks correct though, the actual problem is elsewhere.

The game first create a source voice without pSends (which is correctly linked to mastering voice) and then sets a short infinitely looping buffer which buffer is all zero throughout the game work. Then it creates a number of other source and submix voices. Then upon game exit all those later voices get correctly destroyed by the games. But not that first looping voice. Then games attempts mastering voice deletion but that results in no-op due to the blamed commit (mastering voice is held by that first source voice). Right after that the game frees the buffer data for that first voice (through its own allocator) and then the data get reused somehow, and the new garbage data in the buffer plays in a loop as the concerned "beep" until the game actually exits (that normally takes less than a second but the beep is usually clearly heard, although it loudness depends on the semi-random data in the buffer used after free).

I figured releasing mastering voice affects nothing on Windows as well, hack-skipping that call in game code doesn't reproduce the beep while the buffer is still getting filled with garbage. I also unit-tested with Wine tests that releasing mastering voice specifically (my initial tests for blamed commit were for submix voices only) doesn't really do anything and doesn't stop looping audio.

What actually matters is that the games release audio engine after releasing all those buffers. And that results in destroying everything on Windows. Unit-testing audio engine refcount show that it is not increased when buffers are created. So releasing audio engine even with some leftover buffer tears down everything on Windows. That is consistent with what MS docs say about IXAudio2::Release [1]: "When the final Release is called on a given XAudio2 object, all voice objects that are associated with it are destroyed. Any pointers to these objects that are still held by the client become invalid immediately. Any calls that are made to them cause undefined behavior. ".

I looked at the other usages of voices and audio engine refcounting in FACT.c and XNA_song.c and I think it should be fine with these changes as well (as they don't seem to depend on a voice refcounting engine). But probably another pair on eyes on this won't hurt.

I am attaching a patch with unit test for Wine xaudio2 test suite (I am going to send that Wine upstream once this refcounting is fixed in FAudio and that gets pulled to Wine).

1. https://learn.microsoft.com/en-us/windows/win32/api/xaudio2/nf-xaudio2-ixaudio2-release

[test.txt](https://github.com/user-attachments/files/20359160/test.txt)

### DIFF
--- a/src/FAudio.c
+++ b/src/FAudio.c
@@ -2314,21 +2314,9 @@ static uint32_t check_for_sends_to_voice(FAudioVoice *voice)
 	return ret;
 }
 
-uint32_t FAudioVoice_DestroyVoiceSafeEXT(FAudioVoice *voice)
+static void destroy_voice(FAudioVoice *voice)
 {
-	uint32_t i, ret;
-	LOG_API_ENTER(voice->audio)
-
-	if ((ret = check_for_sends_to_voice(voice)))
-	{
-		LOG_ERROR(
-			voice->audio,
-			"Voice %p is an output for other voice(s)",
-			voice
-		)
-		LOG_API_EXIT(voice->audio)
-		return ret;
-	}
+	uint32_t i;
 
 	/* TODO: Check for dependencies and remove from audio graph first! */
 	FAudio_OPERATIONSET_ClearAllForVoice(voice);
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Huh, I'm surprised to see XAudio2 didn't do deeper refcounting, but XACT doesn't do that either so it makes sense I guess - thanks for catching this!

gofman:
> Huh, I'm surprised to see XAudio2 didn't do deeper refcounting,

In theory it could have private references, e. g., from those voices, but doesn't look like it: attempting to use those objects after releasing the last public reference to engine crashes on Windows, like it is indeed destroying everything as docs suggest.

PR #385, opened 2025-08-22
build: separate WMADEC from PLATFORM_WIN32

rkitover:
Add a new CMake option for WIN32 `WMA_DECODING` to build the WMA codec
decoding code, defaulting to `OFF`.

This makes it independent of `PLATFORM_WIN32` allowing it to be used
with SDL.


flibitijibibo:
Even though it's for SDL this is probably for the Wine maintainers to check out... @rbernon, @gofman, do you know if anyone has time to check this out?

(In the meantime: Note that the WMA support generally should _not_ be used, as it makes portability difficult at the content level which is sort of the opposite of what FNA is for - this could change after the patent portfolio expires though!)

flibitijibibo:
XNA_Song is an "internal" API used by FNA for MediaPlayer support - happily recommend omitting it for basically every other use case!

PR #386, opened 2025-08-23
build: fix warnings for MSVC

rkitover:
Fix some warnings for MSVC. Mostly by type casts. Convert one `fopen()` call to `fopen_s()` with error handling and one double -> int conversion for a buffer size with `ceil()`.

### DIFF
--- a/src/FAudio_platform_sdl3.c
+++ b/src/FAudio_platform_sdl3.c
@@ -61,9 +61,9 @@ static void FAudio_INTERNAL_MixCallback(
 		SDL_PutAudioStreamData(
 			stream,
 			dev->stagingBuffer,
-			dev->stagingLen
+			(int)dev->stagingLen
 		);
-		additional_amount -= dev->stagingLen;
+		additional_amount -= (int)dev->stagingLen;
 	}
 }
 
@@ -265,7 +265,7 @@ uint32_t FAudio_PlatformGetDeviceDetails(
 	FAudio_zero(details, sizeof(FAudioDeviceDetails));
 
 	devs = SDL_GetAudioPlaybackDevices(&devcount);
-	if (index > devcount)
+	if ((int)index > devcount)
 	{
 		SDL_free(devs);
 		return FAUDIO_E_INVALID_CALL;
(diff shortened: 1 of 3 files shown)


src/FAudio_platform_sdl3.c
```diff
@@ -424,7 +424,7 @@ void FAudio_sleep(uint32_t ms)
 
 uint32_t FAudio_timems()
 {
-	return SDL_GetTicks();
+	return (uint32_t)SDL_GetTicks();
```
flibitijibibo, line 427: A good reminder that I need to bump this return value to 64 now 😅

src/FAudio_platform_sdl3.c
```diff
@@ -265,7 +265,7 @@ uint32_t FAudio_PlatformGetDeviceDetails(
 	FAudio_zero(details, sizeof(FAudioDeviceDetails));
 
 	devs = SDL_GetAudioPlaybackDevices(&devcount);
-	if (index > devcount)
+	if (index > (uint32_t)devcount)
```
flibitijibibo, line 268: I worry about the latter being -1, so
```suggestion
	if ((int)index > devcount)
```

src/FAudio_platform_win32.c
```diff
@@ -474,7 +476,7 @@ void FAudio_PlatformInit(
 	args->events[0] = audioEvent;
 	args->events[1] = data->stopEvent;
 	args->audio = audio;
-	if (flags & FAUDIO_1024_QUANTUM) args->updateSize = args->format.Format.nSamplesPerSec / (1000.0 / (64.0 / 3.0));
+	if (flags & FAUDIO_1024_QUANTUM) args->updateSize = (UINT)ceil(args->format.Format.nSamplesPerSec / (1000.0 / (64.0 / 3.0)));
```
flibitijibibo, line 479: This changes behavior, if it should be ceil'd we can do this in another commit

flibitijibibo (changes requested):
Found a couple spots that might change behavior, otherwise good

src/FAudio_platform_win32.c
rkitover: I just thought it makes sense to use `ceil()` for a floating point calculation yielding a buffer size, but it probably does not matter.

flibitijibibo (approved):

PR #392, opened 2025-10-30
Set default 7.1 channel mask to SDL3 and Windows defaults

ivyl:
https://wiki.libsdl.org/SDL3/CategoryAudio#channel-layouts:
  8 channels (7.1) layout: FL, FR, FC, LFE, BL, BR, SL, SR

https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/header-file-changes:

  KSAUDIO_SPEAKER_7POINT1_SURROUND FL, FR, FC, LFE, BL, BR, SL, SR

  In Windows Vista and later versions of Windows, the
  KSAUDIO_SPEAKER_7POINT1 speaker configuration is no longer supported


flibitijibibo:
All makes sense to me, CC me for the SDL side issue whenever that's ready!

ivyl:
I'm working on Pull Request that will fix this. Should be ready in a bit :-)

PR #393, opened 2025-11-21
Patches from Zeb

flibitijibibo:
Just running this in PR form to see if CI has any complaints!

### DIFF
--- a/utils/facttool/facttool.cpp
+++ b/utils/facttool/facttool.cpp
@@ -570,7 +570,7 @@ void FAudioTool_Update()
 				);
 				ImGui::Text(
 					"RPC Code Count: %d",
-					soundBanks[i]->sounds[j].rpcCodeCount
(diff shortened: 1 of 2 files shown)


src/FACT.c
```diff
@@ -659,7 +659,7 @@ uint32_t FACTAudioEngine_CreateStreamingWaveBank(
 	);
 	if (pEngine->prepared_wavebank_count == pEngine->prepared_wavebanks_capacity)
 	{
-		pEngine->prepared_wavebanks_capacity *= 2;
+	    pEngine->prepared_wavebanks_capacity = max(pEngine->prepared_wavebanks_capacity * 2, 8);
```
flibitijibibo: Same but FAudio_max this time

flibitijibibo (commented):
A few review comments for Zeb... a large chunk of this is in the main branch now, it stopped applying cleanly after a couple skips though, will update this in the next version of the changeset.

PR #401, opened 2026-04-13
CI: Run tests with the Sniper debug binaries

flibitijibibo:
Currently fails due to recent regressions, only merge when these are fixed!


flibitijibibo:
Applied this manually, then fixed the tests - main branch should be green again

PR #403, opened 2026-05-10
Fix audio corruption from ADPCM decoding

shaunren:
Fixes an audio static issue for Terraria.

### DIFF
--- a/src/FAudio_internal.c
+++ b/src/FAudio_internal.c
@@ -1911,7 +1911,7 @@ static void decode_mono_adpcm_block(const uint8_t *src, float *dst, uint32_t off
 	src += 7;
 
 	/* Samples */
-	for (i = 0; i < min(2, offset + count); ++i)
+	for (i = 0; i < FAudio_min(2, offset + count); ++i)
 	{
 		if (i < offset) continue;
 		if (i == 0) *dst++ = sample2 / 32768.0;
@@ -1973,7 +1973,7 @@ static void decode_stereo_adpcm_block(const uint8_t *src, float *dst, uint32_t o
 	src += 14;
 
 	/* Samples */
-	for (i = 0; i < min(2, offset + count); ++i)
+	for (i = 0; i < FAudio_min(2, offset + count); ++i)
 	{
 		if (i < offset) continue;
 		if (i == 0)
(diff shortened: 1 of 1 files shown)


src/FAudio_internal.c
```diff
@@ -1911,9 +1911,13 @@ static void decode_mono_adpcm_block(const uint8_t *src, float *dst, uint32_t off
 	src += 7;
 
 	/* Samples */
-	*dst++ = sample2 / 32768.0;
-	*dst++ = sample1 / 32768.0;
-	for (i = 0; i < offset + count; i += 2)
+	for (i = 0; i < min(2, offset + count); ++i)
```
flibitijibibo: Can be FAudio_min instead of min. I wonder if it's worth moving that just above the for as well, just in case it's doing that math every loop...

flibitijibibo (changes requested):
Couple small things, otherwise makes sense... I do wonder if there's a better way to get these offset checks out of the loops but that's probably how this bug happened in the first place.

src/FAudio_internal.c
flibitijibibo:
```suggestion
	for (i = 0; i < FAudio_min(2, offset + count); ++i)
```

src/FAudio_internal.c
```diff
@@ -1968,12 +1973,21 @@ static void decode_stereo_adpcm_block(const uint8_t *src, float *dst, uint32_t o
 	src += 14;
 
 	/* Samples */
-	*dst++ = l_sample2 / 32768.0;
-	*dst++ = r_sample2 / 32768.0;
-	*dst++ = l_sample1 / 32768.0;
-	*dst++ = r_sample1 / 32768.0;
-
-	for (i = 0; i < offset + count; ++i)
+	for (i = 0; i < min(2, offset + count); ++i)
```
flibitijibibo:
```suggestion
	for (i = 0; i < FAudio_min(2, offset + count); ++i)
```

flibitijibibo (commented):
Pushing some buildfixes...

src/FAudio_internal.c
shaunren: Yeah, I'm not sure if it's necessary, since nowadays any decent compiler should be able to optimize it out of the loop.

Issue #362, opened 2024-10-27
Build fails on macOS 15: `ld: pointer not aligned in '_F3DAudioCalculate.lpfReverbDefault'`


barracuda156:
Cannot test this personally, but it fails on CI and also reported in https://trac.macports.org/ticket/70752

```
[ 91%] Linking C shared library libFAudio.dylib
  /opt/local/bin/cmake -E cmake_link_script CMakeFiles/FAudio-shared.dir/link.txt --verbose=ON
[...]

PR #412, opened 2026-09-06
Fix static structs alignment

AlienCowEatCake:
https://github.com/FNA-XNA/FAudio/issues/362

- [x] I confirm that I am the author of this code and release it to the FNA project under the Zlib license. This contribution does not contain code from other sources, including code generated by a Large Language Model ("AI").


flibitijibibo:
Should we document why the alignment has to explicitly be 8? With that explained this should be good to merge

AlienCowEatCake:
Looks like 8 is strictly required only for `F3DAUDIO_DISTANCE_CURVE`, 4 should be enough for `F3DAUDIO_DISTANCE_CURVE_POINT`. I fixed it and added comments

